PR #53, opened 2021-02-15
Remove vendor directory

0x2b3bfa0:
When using [Go modules][1], the `vendor` directory shouldn't be necessary. Given that we're also building and distributing binaries, we probably shouldn't aim to provide support for Go < 1.11.

_Note: this pull request [works on my machine][2] and runs on the GitHub Actions release workflow with [GoReleaser][3]; please verify before merging._

[1]: https://golang.org/doc/go1.11#modules
[2]: https://images-na.ssl-images-amazon.com/images/I/51I1Tcg3B+L.jpg
[3]: https://goreleaser.com/


DavidGOrtega:
@0x2b3bfa0 
> Note: this pull request works on my machine and runs on the GitHub Actions release workflow with GoReleaser, but please verify before merging.

How have you tried this?

0x2b3bfa0:
As previously stated, both locally and here: https://github.com/vgrengvir/terraform-provider-iterative/runs/1902377074

DavidGOrtega:
I can't make it work locally. Probably due to the namespace.
Aside that Im not sure If its a good idea to remove the vendor.  This way for sure the code is never broken and the major con that is the size its not big deal since its not that the case.

0x2b3bfa0:
@DavidGOrtega, the advantage of Go modules is that all the referenced code is pinned to a specific version and a cryptographic hash stored on the `go.sum` file, and all the modules are being cached through [proxy.golang.org](https://proxy.golang.org) just in case a package ceases to exist on its original location. I can't imagine any plausible scenario where code could break because of this.

I would strongly recommend you to drop the `vendor` directory from the code base, and I presume that my admired @FiloSottile would wholeheartedly agree.

0x2b3bfa0:
[Tested and working for 0.13+](https://github.com/0x2b3bfa0/vendor-removal-test/runs/2059593487), with the mouth-watering benefit of making the [pull request review](https://github.com/iterative/terraform-provider-iterative/pull/79) process considerably easier.

DavidGOrtega (approved):
lgtm

PR #54, opened 2021-02-15
Escape shell variables and arguments

0x2b3bfa0:
Prior to this pull request, setting one of the environment variables to a crafted value would allow arbitrary command execution inside the cloud machine.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,4 +14,5 @@ require (
 	github.com/teris-io/shortid v0.0.0-20171029131806-771a37caa5cf
 	github.com/zclconf/go-cty v1.7.0 // indirect
 	golang.org/x/crypto v0.0.0-20201002170205-7f63de1d35b0
+	gopkg.in/alessio/shellescape.v1 v1.0.0-20170105083845-52074bc9df61 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -849,6 +849,8 @@ google.golang.org/protobuf v1.24.0/go.mod h1:r/3tXBNzIEhYS9I1OUVjXDlt8tc493IdKGj
 google.golang.org/protobuf v1.25.0 h1:Ejskq+SyPohKW+1uil0JJMtmHCgJPJ/qWTxr8qp+R4c=
 google.golang.org/protobuf v1.25.0/go.mod h1:9JNX74DMeImyA3h4bdi1ymwjUzf21/xIlbajtzgsN7c=
 gopkg.in/alecthomas/kingpin.v2 v2.2.6/go.mod h1:FMv+mEhP44yOT+4EoQTLFTRgOQ1FBLkstjWtayDeSgw=
+gopkg.in/alessio/shellescape.v1 v1.0.0-20170105083845-52074bc9df61 h1:8ajkpB4hXVftY5ko905id+dOnmorcS2CHNxxHLLDcFM=
+gopkg.in/alessio/shellescape.v1 v1.0.0-20170105083845-52074bc9df61/go.mod h1:IfMagxm39Ys4ybJrDb7W3Ob8RwxftP0Yy+or/NVz1O8=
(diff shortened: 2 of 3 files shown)


DavidGOrtega:
You mean injection?

0x2b3bfa0:
@DavidGOrtega: obviously, they _wouldn't_, and this issue has a limited security impact on all the scenarios I can think of.

Nevertheless, escaping shell variables and arguments has an obvious advantage: being able to include spaces, newlines and quotes. Kubernetes configuration files make an extesive use of these characters, and this is _The Right Way™_ of safely passing literals on a shell script template.

0x2b3bfa0:
Note: #51 introduced changes on the same file; 7ae662a08d82ef94ec6b55e8c9ea6727e5066003 solves the conflicts.

DavidGOrtega:
@0x2b3bfa0 have you tried this? Whats the main difference between now and then? Also, why is this affecting kube?
Could you write a unit test to prove the output is an expected one?

0x2b3bfa0:
> _[...] have you tried this?_

Of course; submitting untested pull requests is not my style.

> _Whats the main difference between now and then?_

Working versus not working when using an environment variable containing newlines, to put an example.

> _Also, why is this affecting kube?_

Because it's going to require a configuration environment variable with newlines and, potentially, quotation marks.

> Could you write a unit test to prove the output is an expected one?

Of course yes, but given that this project doesn't have any unit tests yet, I would prefer not to mix this on the scope of my current task, and less yet on this pull request.

If you want a quick check to try it yourself, you can set `AWS_SECRET_ACCESS_KEY` or another used variable to something like `\"; base64 /dev/random` and running a test apply.

DavidGOrtega:
> Of course; submitting untested pull requests is not my style.

Speaking about the other resources...

> Because it's going to require a configuration environment variable with newlines and, potentially, quotation marks.

OK but why is not that reflected in that resource and have to be included here? Don't get me wrong but Im still not seeing whats the benefit of escaping something that its not going to contain new lines nor quotation mark.

0x2b3bfa0:
> _Speaking about the other resources..._

I haven't tested the other resources, but I've performed manual tests on the changes and they give the expected results. This change is transparent to the environment variable contents, and the only difference introduced by this pull request is that values will be properly quoted and escaped. _Id est:_ `VARIABLE=VALUE` would become `VARIABLE="VALUE"`, _et cetera_.

> _OK but why is not that reflected in that resource and have to be included here? Don't get me wrong but Im still not seeing whats the benefit of escaping something that its not going to contain new lines nor quotation mark._

I'm inclined to consider that this would be the most adequate location for sanitizing the values, and doing otherwise might introduce further development experience and [potentially] user experience shortcomings in the near future. Moreover, preventive value sanitization is **not** optional from the security standpoint.

> _It's a technical debt that we are fixing right now and no PRs is going probably to be accepted without tests.
Here is not very difficult. Just a test that compares the output expected from the template with the generated one._

I don't think _right now_ is the most appropriate way of introducing unit tests on a differently scoped pull request.

> [_Now is better than never,
> Although never is often better than **right** now._](https://www.python.org/dev/peps/pep-0020/)

Please keep in mind that the previous output won't match the current output, as the strings will be surrounded with quotes. I understand the need of introducing testing on the codebase, but I'm not sure about the practicality of adding the test you propose to this pull request.

0x2b3bfa0:
Closing in favor of https://github.com/iterative/terraform-provider-iterative/pull/56

0x2b3bfa0:
Closing in favor of https://github.com/iterative/terraform-provider-iterative/pull/56

0x2b3bfa0:
I've ended up embracing golden tests, but I would still try to refactor the code so we can try the templating function without keeping four static copies of the same script with different variations. It's useful, indeed, but doesn't make development any easier and is, in my opinion, a bit fragile.

PR #76, opened 2021-03-05
Clone AWS CML machine image to eu-west-2

0x2b3bfa0:
Requested by @btjones-me [on Discord](https://discord.com/channels/485586884165107732/728693131557732403/817195215668248616).

PR #79, opened 2021-03-08
Setup machine if no image present

### LINKED ISSUE #72: AWS AMI not present should fallback to prepare the machine on the fly as Azure does
Coming from [here](https://discord.com/channels/485586884165107732/728693131557732403/817195215668248616)

> Trying out v0.3, - aside from changing region, is there anything else I can do to resolve this? In region eu-west-2 (London). Just running the demo from here: https://github.com/iterative/cml_cloud_case/blob/master/.github/workflows/cml.yaml

DavidGOrtega:
If the image is not meant to be present it prepares the machine with all the packages and drivers.
This could have been resolved at a provider level when the provider fails to have the image but ir enforced me to setup another input (calculated) in the TF provider which looked ugly to me and the clone script needs the array anyway.

closes #72


0x2b3bfa0 (approved):
🎉  Great! I haven't run cloud tests on it, but looks awesome.

iterative/aws/provider.go
```diff
@@ -306,6 +306,9 @@ func awsClient(region string) (*ec2.EC2, error) {
 	return svc, err
 }
 
+//ImageRegions provider available image regions
+var ImageRegions = []string{"us-east-1", "us-east-2", "us-west-1", "us-west-2", "eu-north-1", "eu-west-1"}
```
0x2b3bfa0, line 310:
```suggestion
var ImageRegions = []string{"us-east-1", "us-east-2", "us-west-1", "us-west-2", "eu-north-1", "eu-west-1", "eu-west-2"}
```

0x2b3bfa0: It looks like your branch predates #76

PR #80, opened 2021-03-09
Restrict inbound access to SSH

0x2b3bfa0:
> [_Another query, does the security group have to be fully open to inbound traffic or can I close it down without adverse effects?_](https://discord.com/channels/485586884165107732/728693131557732403/818836901842255964) **(From Discord)**

As this user suggested, we might want to thighten a bit the ingress permissions to avoid exposing ports other than SSH (TCP 22) in order to keep the instances as safe as possible.

### DIFF
--- a/iterative/aws/provider.go
+++ b/iterative/aws/provider.go
@@ -91,7 +91,17 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 		})
 
 		if err == nil {
-			ipPermissions := []*ec2.IpPermission{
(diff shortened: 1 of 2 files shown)


DavidGOrtega:
Should not this be set by config? 🤔 
resource_machine might need other ports open.

0x2b3bfa0:
> _Should not this be set by config?_ 🤔
> _resource_machine might need other ports open._

Yes, that would be aligned with the purpose of the machine resource.
Maybe through nested blocks for flexibility?

PR #82, opened 2021-03-10
Migrate string assertion tests to Contains/NotContains

0x2b3bfa0:
I missed this optimization when reviewing #79, sorry. 🤦   It would be nice to take advantage of `testify/assert` expressiveness for string tests by using the inbuilt `Contains` and `NotContains` functions.


DavidGOrtega (approved):
lgtm! 👍

PR #85, opened 2021-03-15
No SG found due to casing

DavidGOrtega:
Fixes the issue that AWS seems to be creating with uppercase and lowecase when creating and searching security groups.
Additionally warns in case the user has no permissions to create new SG.

Removes ./terraform-provider-iterative

closes https://github.com/iterative/cml/issues/447

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,5 @@ terraform
 
 # Ignore the Go vendor directory, superseded by proxy.golang.org and go.{mod,sum}
 vendor
+
+./terraform-provider-iterative
--- a/iterative/aws/provider.go
+++ b/iterative/aws/provider.go
@@ -142,8 +142,11 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	sgDesc, err := svc.DescribeSecurityGroupsWithContext(ctx, &ec2.DescribeSecurityGroupsInput{
 		Filters: []*ec2.Filter{
 			{
-				Name:   aws.String("group-name"),
-				Values: []*string{aws.String(securityGroup), aws.String(strings.Title(securityGroup))},
+				Name: aws.String("group-name"),
+				Values: []*string{
+					aws.String(securityGroup),
+					aws.String(strings.Title(securityGroup)),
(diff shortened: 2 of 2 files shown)


0x2b3bfa0 (approved):
This solution doesn't cover every edge case, as that would require covering all the formula}) possibilities, but sounds like a sensible compromise. See chef-boneyard/chef-provisioning-aws#452 for more information.

DavidGOrtega:
@0x2b3bfa0 I think that the issue is that we create iterative and sometimes the API creates Iterative. Maybe should be safer to also search for ITERATIVE
I dont think that the API is going to rename this as iTeraTiVe... hopefully!

0x2b3bfa0:
@DavidGOrtega, if the title case name has been generated due to an AWS issue, we would be fine with added title case and lowercase checks. Not even my combinatorially purist imagination could imagine an API that creates AlTeRnAtInG case names. 😄

0x2b3bfa0 (approved):
🎉  Looks good to me!

DavidGOrtega:
Have you notice that I have deleted terraform-provider-iterative ?
I did not do it on purpose 🤔

0x2b3bfa0:
@DavidGOrtega what's that? The repository? 😅

DavidGOrtega:
No, the binary file. For some reason was deleted. I think we dont need it. But we should remove it properly (gitignore, etc...)

0x2b3bfa0:
No worries! That file is an artifact shouldn't be there anyways. I wonder why it was there in the first place.

0x2b3bfa0:
Yes, we should `.gitignore` it.

DavidGOrtega:
Yeah , thats what I said. We should remove it properly

0x2b3bfa0:
@DavidGOrtega, it came from #79... `git blame` on me, that overlooked it after the fatigue of ignoring ~50 vendored files. 🙃

DavidGOrtega:
My fault too! 
It's it ready?

0x2b3bfa0:
So it seems, :shipit:

PR #97, opened 2021-03-24
Parallelize clone-amis.go

0x2b3bfa0:
Each `clone-amis.go` run took eons to process, but now it takes a single eon no matter how many regions we want to support.


DavidGOrtega (approved):
lgtm 🚀

PR #103, opened 2021-04-14
Pass --single option to cloud runners

### LINKED ISSUE #94: Missing access to runner parameters --reuse --single
Those parameters should be able to be set also here

0x2b3bfa0:
This pull request passes the newly introduced `--single` option to the launched cloud runners.

Closes #94, required by https://github.com/iterative/cml/pull/476


DavidGOrtega:
@0x2b3bfa0 is our wait mechanism aware of the error is going to throw if the runner is already in use?

DavidGOrtega:
Another issue is that we need to deploy the machine before knowing this. We cant do anything else now but we should add a ticket as a technical debt.
Ne able to check single without launching the machine.

0x2b3bfa0:
@DavidGOrtega, we're running the uniqueness checks on the `cml-runner` side. Should we move them so they run on the instance?

DavidGOrtega:
@0x2b3bfa0 thats what I say. Right now would be a big effort change so would be better to make a ticket for that change just only to avoid spin a machine.

My concern right now is if the runner fails we are logging this correctly in the provider like the await does.

0x2b3bfa0:
@DavidGOrtega, I suspect that we've unadvertently mixed up `--single` and `--reuse` in the previous comments after reading #94. The former argument is being passed to the `cml-runner` process on the cloud instance (as per this pull request), while the latter is being handled locally on the original non-cloud invocation.

If the runner fails and emits the JSON line we're expecting, the behavior should be similar to any other failure case. 🤔  Is that what you meant?

DavidGOrtega (approved):
lgtm

PR #104, opened 2021-04-29
Debug the instance IP prior to awaiting the runner

0x2b3bfa0:
This would be extremely useful for debugging some connection issues.

### DIFF
--- a/iterative/resource_runner.go
+++ b/iterative/resource_runner.go
@@ -190,7 +190,8 @@ func resourceRunnerCreate(ctx context.Context, d *schema.ResourceData, m interfa
 	if diags.HasError() {
 		return diags
 	}
-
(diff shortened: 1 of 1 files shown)


DavidGOrtega:
@0x2b3bfa0 do you think so?
I mean, that log will only be visible if we setup the debug level as env variable right? Or maybe is there another way?

0x2b3bfa0:
@DavidGOrtega, yes, the nice thing is that we can handle all that through the environment and sedn the debug output to a file for further investigation. Please check [this workflow](https://github.com/iterative/cml/issues/484#issuecomment-828823161) for an example.

DavidGOrtega:
@0x2b3bfa0 got it!

DavidGOrtega (approved):
lgtm

PR #120, opened 2021-05-26
Mark resource_runner token as sensitive

0x2b3bfa0:
Related to https://github.com/iterative/cml/pull/564


DavidGOrtega (approved):
👍 
awesome. It's going to print back another value?

0x2b3bfa0:
It will mask the output of commands like `terraform plan` and `terraform apply` so the value gets replaced, literally, with `(sensitive value)`

Issue #44, opened 2021-01-18
When machine stops public ip changes


DavidGOrtega:
It should be accesible through DNS

DavidGOrtega:
Tried to discover that in AWS the fqdn changes with the ip.
The only way to fix this would be use elastic IPs

0x2b3bfa0:
* Does the public IP change for the same machine once and if it has been restarted?
* How are our machines/runners affected by this issue?

DavidGOrtega:
> Does the public IP change for the same machine once and if it has been restarted?

yes unless you use an elastic ip

0x2b3bfa0:
If we choose not to implement https://github.com/iterative/terraform-provider-iterative/issues/122, would this issue matter at all?

0x2b3bfa0:
To be reopened along with #122 if we choose to implement that feature

PR #129, opened 2021-06-09
Reboot to enable nvidia

DavidGOrtega:
Seems that nvidia needs to be rebooted now. So we are rebooting and enabling cml after reboot, additionally we are adding the latest nvidia ppa.

### DIFF
--- a/iterative/resource_runner.go
+++ b/iterative/resource_runner.go
@@ -270,7 +270,8 @@ sudo apt update && sudo apt-get install -y terraform
 curl -sL https://deb.nodesource.com/setup_12.x | sudo bash
 sudo apt update && sudo apt-get install -y nodejs
 
-sudo apt install ubuntu-drivers-common
+sudo add-apt-repository ppa:graphics-drivers/ppa -y
+sudo apt update && sudo apt install ubuntu-drivers-common
 sudo ubuntu-drivers autoinstall
 curl -s -L https://nvidia.github.io/nvidia-docker/gpgkey | sudo apt-key add -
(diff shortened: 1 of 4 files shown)


DavidGOrtega:
@0x2b3bfa0 ready please test 🙏

DavidGOrtega:
Ok this should not work in aws, however... it does.
We do no use elastic IPs and in a naive test TF is giving me this IP 18.144.XXX.XXX however when I get in the machine it clearly says:

ubuntu@ip-172-31-XXX-XXX which is obviously different, however for some reason it works and sticks...

0x2b3bfa0:
> _Ok this should not work in aws, however... it does._

If you mean because of #44, yes, it wouldn't

> clearly says: `ubuntu@ip-172-31-XXX-XXX` which is obviously different

This is not the public address, but the private one: `172.31.0.0/16` is one of the 16 class B private networks in the "172" block.

DavidGOrtega:
I overlooked the 31 because of #44 seems to be working in AWS too!

DavidGOrtega:
image

0x2b3bfa0:
There are some nuances:
* #44 is about stopping and starting an EC2 instance; that changed, changes, and will change the address each time.
* #129 is about rebooting the operating system without altering the machine state, and that doesn't change the address.

DavidGOrtega:
> There are some nuances:

To me they should be aligned like in Azure but I do not design AWS 😄

0x2b3bfa0 (approved):
Actually, the image boots with `5.4.0-1047-azure` but drivers get installed for `5.4.0-1048-azure` so rebooting is the only way of loading the installed drivers.

Linus, I'm thinking exactly that.

PR #130, opened 2021-06-09
Fix nil pointer dereference on SSH key parser

0x2b3bfa0:
Triggered when feeding with an invalid key

> ```go
> func Decode(data []byte) (p *Block, rest []byte)
> ```
> [_If no PEM data is found, `p` is `nil` and the whole of the input is returned in rest._](https://golang.org/pkg/encoding/pem/#Decode )


DavidGOrtega (approved):
👍

PR #135, opened 2021-06-10
Hotfix/azure only reboot nvidia

DavidGOrtega:
Reboots only Azure. AWS seems to be loosing the public IP randomly (not everytime).
It's flaky and we should find a solution in the long term. Probably:
  - adding elastic ips in AWS 
  - NO having to restart
  - Azure AMIs
  
We leave it for good as we need a hotfix

### DIFF
--- a/iterative/resource_runner.go
+++ b/iterative/resource_runner.go
@@ -345,8 +345,8 @@ sudo reboot
 {{else}}
 sudo systemctl daemon-reload
 sudo systemctl enable cml.service --now
-{{end}}
-{{end}}
+{{- end}}
+{{- end}}
 {{- end}}
 `)
 	var customDataBuffer bytes.Buffer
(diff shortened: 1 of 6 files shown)


iterative/resource_runner.go
```diff
@@ -338,8 +338,14 @@ sudo bash -c 'cat << EOF > /etc/systemd/system/cml.service
...
 EOF'
 
+{{if eq .cloud "azure"}}
 sudo systemctl enable cml.service
 sudo reboot
+{{else}}
+sudo systemctl daemon-reload
+sudo systemctl enable cml.service --now
+{{end}}
+
```
0x2b3bfa0, line 348:
```suggestion
```

Awesome! Please remove the newline or add an hyphen so we don't end the file with too many newlines: @restyled-commits will complain.

0x2b3bfa0 (changes requested):

0x2b3bfa0 (approved):
😅

PR #143, opened 2021-06-15
AWS better check and election

### LINKED ISSUE #99: AMI fallback script is happening with provider mapped regions
If I set a "iterative region" like us-west the full script with CUDA installation is happening

DavidGOrtega:
- Ensures that the Ubuntu image is developed by Canonical
 - Chooses the default VPC and not the very first one
 - Ensures that the subnet can create public IPs and it has available IPs still
 - Checks the iterative mapped regions to determine if ami is available. closes #99 
 - Sets the used ami name in the tfstate

### DIFF
--- a/iterative/aws/provider.go
+++ b/iterative/aws/provider.go
@@ -26,7 +26,7 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	instanceName := d.Get("name").(string)
 	pairName := d.Id()
 	hddSize := d.Get("instance_hdd_size").(int)
-	region := getRegion(d.Get("region").(string))
+	region := GetRegion(d.Get("region").(string))
 	instanceType := getInstanceType(d.Get("instance_type").(string), d.Get("instance_gpu").(string))
 	ami := d.Get("image").(string)
 	keyPublic := d.Get("ssh_public").(string)
@@ -38,7 +38,6 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	}
 
 	svc, err := awsClient(region)
-
 	if err != nil {
 		return decodeAWSError(region, err)
 	}
@@ -69,6 +68,10 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 				{
 					Name:   aws.String("architecture"),
 					Values: []*string{aws.String("x86_64")},
+				},
+				{
+					Name:   aws.String("owner-id"),
+					Values: []*string{aws.String("099720109477")},
 				},
 			},
 		})
@@ -101,10 +104,18 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 		securityGroup = "iterative"
 
 		vpcsDesc, _ := svc.DescribeVpcs(&ec2.DescribeVpcsInput{})
+
 		if len(vpcsDesc.Vpcs) == 0 {
 			return errors.New("no VPCs found")
 		}
 		vpcID = *vpcsDesc.Vpcs[0].VpcId
+
+		for _, vpc := range vpcsDesc.Vpcs {
+			if *vpc.IsDefault {
+				vpcID = *vpc.VpcId
+				break
+			}
+		}
 
 		gpResult, err := svc.CreateSecurityGroup(&ec2.CreateSecurityGroupInput{
 			GroupName:   aws.String(securityGroup),
@@ -177,6 +188,16 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	if len(subDesc.Subnets) == 0 {
 		return errors.New("no subnets found")
 	}
+	var subnetID string
+	for _, subnet := range subDesc.Subnets {
+		if *subnet.AvailableIpAddressCount > 0 && *subnet.MapPublicIpOnLaunch {
+			subnetID = *subnet.SubnetId
+			break
+		}
+	}
+	if subnetID == "" {
+		return errors.New("No subnet found with public IPs available or able to create new public IPs on creation")
+	}
 
 	blockDeviceMappings := []*ec2.BlockDeviceMapping{
 		{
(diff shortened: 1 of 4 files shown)


DavidGOrtega:
> Checks the iterative mapped regions to determine if ami is available.

This should close another issue that I remember we opened @0x2b3bfa0

0x2b3bfa0:
#99?

iterative/resource_runner.go
```diff
@@ -438,8 +438,13 @@ func isAMIAvailable(cloud string, region string) bool {
 		regions = azure.ImageRegions
 	}
 
+	var regionCloud = aws.GetRegion(region)
+	if cloud == "azure" {
+		regionCloud = azure.GetRegion(region)
+	}
```
0x2b3bfa0, line 444:
```suggestion
	var regionCloud string
	switch cloud {
	case "aws":
		regionCloud = aws.GetRegion(region)
	case "azure":
		regionCloud = azure.GetRegion(region)
	}
```

Can we use a `switch` structure so it's cleaner once we add more vendors?

iterative/aws/provider.go
```diff
@@ -169,6 +180,10 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 				Name:   aws.String("vpc-id"),
 				Values: []*string{aws.String(vpcID)},
 			},
+			{
+				Name:   aws.String("available-ip-address-count"),
+				Values: []*string{aws.String(vpcID)},
+			},
```
0x2b3bfa0, line 186: What does this do? After reading [the documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSubnets.html), it appears like you're searching for subnets that have a given number of addresses, but that number seems to be the identifier of the VPC. 🤔  Could it be a copy-paste mistake?

iterative/aws/provider.go
```diff
@@ -70,6 +69,10 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 					Name:   aws.String("architecture"),
 					Values: []*string{aws.String("x86_64")},
 				},
+				{
+					Name:   aws.String("owner-id"),
+					Values: []*string{aws.String("099720109477")},
```
0x2b3bfa0, line 74: 💯  Awesome! Otherwise, it could pick any machine image out of there.

iterative/resource_runner.go
```diff
@@ -438,8 +438,13 @@ func isAMIAvailable(cloud string, region string) bool {
...
+
 	for _, item := range regions {
-		if item == region {
+		if item == regionCloud {
```
0x2b3bfa0, line 447: Any reason to use regionCloud instead of cloudRegion?

iterative/resource_runner.go
DavidGOrtega: sure! I never use them because I hate them. In many languages is so missused and so error prone that  I almost removed it from my voc

0x2b3bfa0: If you prefer, we can keep extending it with chained conditionals. I'm open to any of the approaches.

Go's `switch` statement is pretty sane in my opinion, and doesn't [`fallthrough`](https://github.com/golang/go/wiki/Switch#fall-through) unless you explicitly require it to.

iterative/aws/provider.go
DavidGOrtega: This is not needed and indeed no test done by me was done with this section, if this work we wont have any VPC unless  the id coincides with the number of ips!

DavidGOrtega: I think that VScode added that while saving from a previous atempt? No clue

0x2b3bfa0:
```suggestion
```

I barely can't refrain myself from creating a new [humorous unit of measurement](https://en.wikipedia.org/wiki/List_of_humorous_units_of_measurement), the MiracleBlock.

iterative/resource_runner.go
DavidGOrtega: Ideally we should write something like:


```go
var provider Provider
	switch cloud {
		case "aws":
		provider = aws
	case "azure":
		provider = azure
	}

	regions := provider.ImageRegions
	cloudRegion := provider.GetRegion(region)
	for _, item := range regions {
		if item == cloudRegion {
			return true
		}
	}

	return false
```

But I don't know how to do Interfaces in go

0x2b3bfa0: :100: I have been thinking on this for a few months, but we always had something more important to do. If you want, I can open a follow-up pull request with a proposal similar to your snippet above.

0x2b3bfa0: Closing this thread with 436bfa209dca1fed319ef38cd639d1d1e875bed7 and #147

iterative/resource_runner.go
```diff
@@ -438,8 +438,16 @@ func isAMIAvailable(cloud string, region string) bool {
...
 	}
 
+	var regionCloud string
+	switch cloud {
+	case "aws":
+		regionCloud = aws.GetRegion(region)
+	case "azure":
+		regionCloud = azure.GetRegion(region)
+	}
+
```
0x2b3bfa0, line 448:
```suggestion
	var cloudRegion string
	switch cloud {
	case "aws":
		cloudRegion = aws.GetRegion(region)
	case "azure":
		cloudRegion = azure.GetRegion(region)
	}

```

0x2b3bfa0 (approved):
Looks good to me!

iterative/resource_runner_test.go
```diff
@@ -16,31 +16,53 @@ func TestScript(t *testing.T) {
...
 	})
+	
+	t.Run("AWS known generic region should add the NVIDA drivers", func(t *testing.T) {
+		data := make(map[string]interface{})
+		data["ami"] = isAMIAvailable("aws", "us-west")
+
+		script, err := renderScript(data)
+		assert.Nil(t, err)
+		assert.Contains(t, script, "sudo ubuntu-drivers autoinstall")
+	})
```
0x2b3bfa0, line 31:
```suggestion
	t.Run("AWS known generic region should not add the NVIDA drivers", func(t *testing.T) {
		data := make(map[string]interface{})
		data["ami"] = isAMIAvailable("aws", "us-west")

		script, err := renderScript(data)
		assert.Nil(t, err)
		assert.NotContains(t, script, "sudo ubuntu-drivers autoinstall")
	})
```

0x2b3bfa0 (approved):
Now with tests for #99

casperdcl (approved):
tests aren't exactly very DRY but whatever

Issue #118, opened 2021-05-26
Azure nvidia drivers seems to not be installed properly


DavidGOrtega:
Comming from [here](https://discord.com/channels/485586884165107732/728693131557732403/847038440612298752)

PR #161, opened 2021-07-09
Wait improvements

### LINKED ISSUE #153: Reduce ssh heartbeat time
Currently it does it every 10 seconds which is an enormous amount of time

```go
return utils.RunCommand("journalctl --no-pager", 10*time.Second,
```

### LINKED ISSUE #154: The error log is so long that is completely useless
Im having an issue starting an AWS machine and the log is so ridiculously long that is not handled in the shell completely and would take me more time to read it than read The Bible in latin.

DavidGOrtega:
- fixes #154 Only displays the service logs 
  - fixes #153 Reduces the timeout from 10 to 2

### DIFF
--- a/iterative/resource_runner.go
+++ b/iterative/resource_runner.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"os"
 	"strconv"
 	"text/template"
@@ -210,7 +211,17 @@ func resourceRunnerCreate(ctx context.Context, d *schema.ResourceData, m interfa
 	var logError error
 	var logEvents string
 	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
-		logEvents, logError = resourceMachineLogs(ctx, d, m)
+		switch cloud := d.Get("cloud").(string); cloud {
+		case "kubernetes":
+			logEvents, logError = resourceMachineLogs(ctx, d, m)
+		default:
+			logEvents, logError = utils.RunCommand("journalctl --unit cml --no-pager",
+				2*time.Second,
+				net.JoinHostPort(d.Get("instance_ip").(string), "22"),
+				"ubuntu",
+				d.Get("ssh_private").(string))
+		}
+
 		log.Printf("[DEBUG] Collected log events: %#v", logEvents)
 		log.Printf("[DEBUG] Connection errors: %#v", logError)
 
--- a/iterative/aws/provider.go
+++ b/iterative/aws/provider.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
 	"regexp"
 	"sort"
 	"strconv"
@@ -16,8 +15,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
-	"terraform-provider-iterative/iterative/utils"
 )
 
 //ResourceMachineCreate creates AWS instance
@@ -439,8 +436,4 @@ func decodeAWSError(region string, err error) error {
 	}
 
 	return fmt.Errorf("Not able to decode: %s", err.Error())
-}
-
-func ResourceMachineLogs(ctx context.Context, d *schema.ResourceData, m interface{}) (string, error) {
-	return utils.RunCommand("journalctl --no-pager", 10*time.Second, net.JoinHostPort(d.Get("instance_ip").(string), "22"), "ubuntu", d.Get("ssh_private").(string))
 }
--- a/iterative/azure/provider.go
+++ b/iterative/azure/provider.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"strings"
 	"time"
@@ -17,8 +16,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
-	"terraform-provider-iterative/iterative/utils"
 )
 
 //ResourceMachineCreate creates AWS instance
@@ -395,8 +392,4 @@ func subscriptionID() (string, error) {
 	}
 
 	return "", errors.New("AZURE_SUBSCRIPTION_ID is not present")
-}
-
-func ResourceMachineLogs(ctx context.Context, d *schema.ResourceData, m interface{}) (string, error) {
-	return utils.RunCommand("journalctl --no-pager", 10*time.Second, net.JoinHostPort(d.Get("instance_ip").(string), "22"), "ubuntu", d.Get("ssh_private").(string))
 }
--- a/iterative/resource_machine.go
+++ b/iterative/resource_machine.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"net"
 	"time"
 
 	"terraform-provider-iterative/iterative/aws"
@@ -236,13 +237,13 @@ func resourceMachineRead(ctx context.Context, d *schema.ResourceData, m interfac
 
 func resourceMachineLogs(ctx context.Context, d *schema.ResourceData, m interface{}) (string, error) {
 	switch cloud := d.Get("cloud").(string); cloud {
-	case "aws":
-		return aws.ResourceMachineLogs(ctx, d, m)
-	case "azure":
-		return azure.ResourceMachineLogs(ctx, d, m)
 	case "kubernetes":
 		return kubernetes.ResourceMachineLogs(ctx, d, m)
 	default:
-		return "", fmt.Errorf("Unknown cloud: %s", cloud)
+		return utils.RunCommand("journalctl --no-pager",
+			2*time.Second,
+			net.JoinHostPort(d.Get("instance_ip").(string), "22"),
+			"ubuntu",
+			d.Get("ssh_private").(string))
 	}
 }


0x2b3bfa0 (commented):
* #153

Reducing the connection timeout from 10 to 2 seconds in the call to `runCommand` may not have the desired effect. It will only limit the timeout of each connection, but not the delay between tries, which is being [automatically handled](https://github.com/hashicorp/terraform-plugin-sdk/blob/2387aca0b8a1347c17b36b826c0f56d9258df457/helper/resource/state.go#L182-L198) by Terraform's exponential backoff algorithm. If you want to reduce the latter, we would need to use the raw [`WaitForStateContext`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#StateChangeConf.WaitForStateContext) instead. Is this change worth the maintenance overhead? I don't think people will notice the 10 second difference.

* #154

In my opinion, the full instance log is **not** competely useless. It will only show in case of failure, and it has helped us in the past to diagnose quite a few difficult issues. Restricting the log output to the `cml` unit when running on raw machines (as opposed to orchestrators) would leave us without any easy way of retrieving the whole low-level logs in case of failure. Addressing your concerns from #154, you can always search for `cml` in the whole logs to get the runner-specific lines.

iterative/resource_runner.go
```diff
@@ -210,7 +211,17 @@ func resourceRunnerCreate(ctx context.Context, d *schema.ResourceData, m interfa
...
+		case "kubernetes":
+			logEvents, logError = resourceMachineLogs(ctx, d, m)
+		default:
+			logEvents, logError = utils.RunCommand("journalctl --unit cml --no-pager",
+				2*time.Second,
+				net.JoinHostPort(d.Get("instance_ip").(string), "22"),
+				"ubuntu",
+				d.Get("ssh_private").(string))
+		}
+
```
0x2b3bfa0, line 224: I'd prefer to keep a separate `resourceMachineLogs` function for each provider, even if some of them share the same code. The proposed changes are introducing two additional calls for non-SSH vendors, and it doesn't feel quite elegant, in my opinion.

DavidGOrtega:
> In my opinion, the full instance log is not competely useless. It will only show in case of failure, and it has helped us in the past to diagnose quite a few difficult issues. Restricting the log output to the cml unit when running on raw machines (as opposed to orchestrators) would leave us without any easy way of retrieving the whole low-level logs in case of failure. Addressing your concerns from #154, you can always search for cml in the whole logs to get the runner-specific lines.

Its not restricting the machine logs its restricting the runner logs and to be honest I do not see current logs useful at all.

> Reducing the connection timeout from 10 to 2 seconds in the call to runCommand may not have the desired effect. It will only limit the timeout of each connection, but not the delay between tries, which is being automatically handled by Terraform's exponential backoff algorithm. If you want to reduce the latter, we would need to use the raw WaitForStateContext instead. Is this change worth the maintenance overhead? I don't think people will notice the 10 second difference.

I think it's totally correct. Your approach is try to connect via SSH with 10 seconds timeout. I prefer to do it several short times because the runner will be destroyed after 20 secs aprox having, the 10 secs screen approach, to have tried twice at most. The problem is that the SSH connection timeout does not seems to be effective in terms of being able to connect with those 10 secs once the connection is effective.

A batch of 60 runs with both methods shows empirically that this method succeeded always while the other failed multiple times  showing an error in line 14 of utils.js (executing terraform)

iterative/resource_runner.go
DavidGOrtega: IMHO having duplicate code is not elegant at all

DavidGOrtega: And of course we are having the resource machine which can have intermediate business logic

DavidGOrtega: > The proposed changes are introducing two additional calls for non-SSH vendors

Which two calls?

0x2b3bfa0:
> _I think it's totally correct. Your approach is try to connect via SSH with 10 seconds timeout. I prefer to do it several short times because the runner will be destroyed after 20 secs aprox having, the 10 secs screen approach, to have tried twice at most. The problem is that the SSH connection timeout does not seems to be effective in terms of being able to connect with those 10 secs once the connection is effective._
>
> _A batch of 60 runs with both methods shows empirically that this method succeeded always while the other failed multiple times showing an error in line 14 of utils.js (executing terraform)_

If the old 10 second timeout had a lower success rate than the proposed 2 second timeout, it's definitely worth the change. As long as it has time to establish the connection and retrieve a few thousand lines in the worst case, everything should be fine.

DavidGOrtega:
> As long as it has time to establish the connection and retrieve a few thousand lines in the worst case, 

timeout is not for execution, timeout is for connection

DavidGOrtega:
Also regarding to logs is that I do not need more logs of what I would see if I were launching a runner locally in my machine (non cloud)

0x2b3bfa0:
> _timeout is not for execution, timeout is for connection_

True! I thought that we were accounting for both, but I only specified the [`ClientConfig`](https://pkg.go.dev/golang.org/x/crypto/ssh#ClientConfig) one. 🤦🏼

DavidGOrtega:
Indeed makes no sense to make it for all. This way we can connect within 2 secs and can still consume 18 secs of logs

0x2b3bfa0:
> _Its not restricting the machine logs its restricting the runner logs_ [...]

Indeed, you're restricting the `resource_cml_runner` logs, not the `resource_machine` logs. Incidentally, the former are the only logs that we ever print on error; the latter won't be accessible by any means. 🙊 

> [...] _and to be honest I do not see current logs useful at all._
> _Also regarding to logs is that I do not need more logs of what I would see if I were launching a runner locally in my machine (non cloud)_

I still disagree on this. When launching a runner locally in your machine, you're in control of your environment and know if your system is working correctly and how to debug it. When launching a runner in an ephemeral cloud instance that gets termianted on error, it would make sense to have a way of retrieving the full logs before they get lost. I don't care if it's hidden behind a debug environment variable, but there **shoud** be a way to do that, lest spending hours of trial and error debugging some machine-level issues.

DavidGOrtega:
> the latter won't be accessible by any means.

Exactly what we need. We just only need to provide errors from the cloud (at creation) and runner. The rest is users responsibility (as the vendors do). Anyway Im not the only one who thinks that our logs are too much

DavidGOrtega:
The logs for the plain machine can be displayed inside the machine resource code. WE can open a ticket for that and thats the reason I have not removed the method wich is indeed useful

0x2b3bfa0:
> _Exactly what we need. We just only need to provide errors from the cloud (at creation) and runner. The rest is users responsibility (as the vendors do)._

Unlike vendors, we offer custom machine images that fail from time to time and produce machine-level logs. Are we supposed to ignore errors produced at the machine level and make users responsible for them? I see some cases where that wouldn't be convenient; see #118 for an example.

> _Anyway Im not the only one who thinks that our logs are too much_

If there is more people thinking the same and they have GitHub accounts, it would be nice to bring them into this conversation so we can get a richer context.

> _The logs for the plain machine can be displayed inside the machine resource code._

Given that the `resource_machine` gets completed as soon as the machine is up and running, we wouldn't be able to print any meaningful log before Terraform exits.

> _WE can open a ticket for that and thats the reason I have not removed the method wich is indeed useful_

Agreed! What if we always print the full logs for `resource_cml_runner` machines to the Terraform debug output (not visible by default) and the runner-specific logs to the standard output just in case of error?

iterative/resource_runner.go
0x2b3bfa0: > _IMHO having duplicate code is not elegant at all_

Agreed! Nevertheless, in this case we're just duplicating a line of code for the sake of having the same methods on every provider. We aren't copying and pasting a copy of the `utils.RunCommand` function, but just a single line call. 🙃 

> _And of course we are having the resource machine which can have intermediate business logic_

Intermediate busniess logic is a pretty general term here. While 66% of our supported vendors — or 75% very soon — are going to rely on SSH for log gathering, that percentage will considerably decrease as soon as we start adding more container orchestrators. Should we count a specific log gathering mechanism as _intemediate business logic_ just because it's the same for a couple of providers?

DavidGOrtega:
> Agreed! What if we always print the full logs for resource_cml_runner machines to the Terraform debug output (not visible by default) and the runner-specific logs to the standard output just in case of error?

Sounds good! Can you please open a ticket? We need to sort out this one because as I said its failing the deployment sometimes.

iterative/resource_runner.go
0x2b3bfa0: > _Which two calls?_

For orchestrators and other non-SSH providers, we would be calling the following two blocks sequentially, hitting two identical `switch` statements (hello code duplication) and making two calls: one to `resourceMachineLogs` and other to `provider.ResourceMachineLogs`

https://github.com/iterative/terraform-provider-iterative/blob/05aa3a385ea295a6d3707441a7fb6651f2347fc7/iterative/resource_runner.go#L214-L223
https://github.com/iterative/terraform-provider-iterative/blob/05aa3a385ea295a6d3707441a7fb6651f2347fc7/iterative/resource_machine.go#L238-L249

0x2b3bfa0:
> _Sounds good! Can you please open a ticket? We need to sort out this one because as I said its failing the deployment sometimes._

Better yet, can you open a separate pull request for the SSH connection timeout and leave this for the log along with #154?

DavidGOrtega:
> Better yet, can you open a separate pull request for the SSH connection timeout and leave this for the log along with #154?

No, I wont say its better.
That enormous amount of logs is not yet useful for the runner to justify it

DavidGOrtega:
open #163

DavidGOrtega:
@0x2b3bfa0 we need to merge this


```
 Jul 09 14:31:16 ip-172-31-11-183 systemd[5759]: Listening on GnuPG
│ cryptographic agent and passphrase cache (access for web browsers).
│ Jul 09 14:31:16 ip-172-31-11-183 systemd[5759]: Reached target Sockets.
│ Jul 09 14:31:16 ip-172-31-11-183 systemd[5759]: Reached target Basic
│ System.
│ Jul 09 14:31:16 ip-172-31-11-183 systemd[5759]: Reached target Default.
│ Jul 09 14:31:16 ip-172-31-11-183 systemd[5759]: Startup finished in 19ms.
│ Jul 09 14:31:16 ip-172-31-11-183 systemd[1]: Started User Manager for UID
│ 1000.
│ 
╵
    at /usr/lib/node_modules/@dvcorg/cml/src/utils.js:14:27
    at ChildProcess.exithandler (child_process.js:315:5)
    at ChildProcess.emit (events.js:314:20)
    at maybeClose (internal/child_process.js:1022:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:287:5)
Cleaning up file based variables
00:00
ERROR: Job failed: exit code 1
```

0x2b3bfa0:
> _we need to merge this_

Then, let's merge it we need to. I'm not giving an explicit high-level approval and definitely would prefer to separate #153 from #154, but yes, it works, and availability goes first.

0x2b3bfa0:
@DavidGOrtega, I'm curious to see the full log you quote on https://github.com/iterative/terraform-provider-iterative/pull/161#issuecomment-877255689. Can you please attach it?

Issue #185, opened 2021-09-05
Support for cloud resource metadata


0x2b3bfa0:
### Epic for iterative/cml#430 provider-side support; follow-up to #3

Some vendors rely heavily on key/value metadata for resource management and, in some specific cases, permissions and policy enforcing. Most production–grade policies on vendors like AWS won't allow users to manage resources without proper creation–time tags.

Moreover, data science teams might also be interested in adding metadata to their cloud experiments for cost management and traceability purposes. See [Kubeflow Metadata](https://v0-7.kubeflow.org/docs/components/misc/metadata/) for a similar concept.

### Implementation

- [x] https://github.com/iterative/terraform-provider-iterative/pull/193
  - [x] https://github.com/iterative/tpi/issues/10
  - [x] https://github.com/iterative/terraform-provider-iterative/pull/187, depends on https://github.com/iterative/terraform-provider-iterative/pull/183
  - [x] https://github.com/iterative/terraform-provider-iterative/pull/191
  - [x] https://github.com/iterative/terraform-provider-iterative/pull/190
  - [x] https://github.com/iterative/terraform-provider-iterative/pull/189

#### Naming

* **`metadata`** (subject to change)
* `labels`
* `tags`

#### Typing

* `map[string]string`

#### Defaults

The official AWS [provider](https://registry.terraform.io/providers/hashicorp/aws) supports [default tags](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-configuration-block) for all the created resources, but implementing this approach in a multicloud setting doesn't look like a good idea: every vendor uses labels/tags in a different way, and providing sensible defaults could be very difficult.

* https://github.com/iterative/terraform-provider-iterative/pull/187#issue-727345841

#### Example

```hcl2
resource "iterative_machine" "machine" {
  ···
  metadata = {
    project = "example"
    creator = "anybody"
  }
}
```

### Requests

* [Slack](https://iterativeai.slack.com/archives/C01R00PPQ1L/p1630113902003100?thread_ts=1630088323.001600&cid=C01R00PPQ1L) 
* [Discord](https://discord.com/channels/485586884165107732/728693131557732403/882053329193168906)

DavidGOrtega:
@0x2b3bfa0  Are we missing docs?

0x2b3bfa0:
@DavidGOrtega, yes, we're missing documentation and API propagation to:
* iterative/cml#430 
[...]

PR #190, opened 2021-09-05
Add support for Google Cloud metadata (labels)

0x2b3bfa0:
Part of https://github.com/iterative/terraform-provider-iterative/issues/185


DavidGOrtega (approved):
🚀 lgtm!

0x2b3bfa0:
Note for the future: there is [another kind of resource-level tags](https://cloud.google.com/resource-manager/docs/tags/tags-creating-and-managing) more similar to other vendors', but I won't implement that if users don't ask.

PR #203, opened 2021-09-10
Improve Makefile

0x2b3bfa0:
I don't use `darwin_amd64` to develop the provider, and having to change the hardcoded values each time I checkout the project is really painful.

Now testing the provider is as easy as running `make` in any platform.

* Remove hardcoded references to GOOS and GOARCH
* Use `go install` for the `install` target
* Unify parentheses versus curly braces
* Abstract duplicate path


DavidGOrtega (approved):
🚀 ❤️ lgtm!

PR #205, opened 2021-09-11
Fix ineffective sudo usage in setup.sh

0x2b3bfa0:
Overlooked when reviewing #170, `sudo` is being applied to `echo` but the redirection operator still has the permissions of the main process.


DavidGOrtega (approved):
👍 lgtm

Issue #197, opened 2021-09-07
Azure machine images


0x2b3bfa0:
### Quote from https://github.com/iterative/cml/issues/162#issuecomment-792704409

> Currently we are recreating from scratch the instance environment each time we run a continuous integration job on Azure, because [(apparently)](https://docs.microsoft.com/en-us/azure/marketplace/marketplace-virtual-machines) we can't share instances images with a wider audience through the Azure Marketplace without having the [Microsoft Partner](https://docs.microsoft.com/en-us/partner-center/mpn-create-a-partner-center-account) qualification. (?)
> 
> The current workaround implies significantly longer total job execution times due to the increased instance configuration times, so being able to publish prebuilt CML base images to the Azure Marketplace —as we're already doing for AWS machine images— could be a nice improvement for end users.

0x2b3bfa0:
Dealt with as part of #196; no need of a separate issue unless we plan to publish markeplace images.

Issue #196, opened 2021-09-07
Machine images


0x2b3bfa0:
* [x] Merge https://github.com/iterative/terraform-provider-iterative/pull/409
* [x] Merge https://github.com/iterative/terraform-provider-iterative/pull/384
* [x] https://github.com/iterative/terraform-provider-iterative/issues/410

### ~Follow-up on https://github.com/iterative/cml/issues/162~

- [X] ~AWS~
- [ ] ~#197~
- [ ] ~GCP: see https://cloud.google.com/compute/docs/images/managing-access-custom-images#share-images-publicly~
- [X] ~Kubernetes ([container images on iterative/cml](https://github.com/iterative/cml/blob/899c6d2e3eaffc66d12492deffa658e8430f1c66/Dockerfile))~

0x2b3bfa0:
Moved remaining itmes to #562

PR #206, opened 2021-09-12
Upgrade machine image build system

0x2b3bfa0:
This pull request was originally meant to publish our custom machine images on all the supported cloud providers. It has been repurposed to commit into this repository what we already have in production.

### Outdated checklist

* [x] ~Get rid of [`clone-amis.go`](https://github.com/iterative/terraform-provider-iterative/blob/8d6f803f1d02e152e7bb8dc39a006a3466ed91f8/cml/clone-amis.go) in favor of Packer's inbuilt multi-region support~
* [x] ~Migrate from JSON to HCL2, the new recommended format for Packer configuration files~
* [x] ~Tag all as many AWS resources as possible and use them for the Packer role policies~
* [x] ~Let Packer delete all the old images and leave only the latest one for each region~
* [x] ~Implement experimental support for Azure and Google Cloud~
* [x] ~Create a GitHub Actions workflow for testing and deploying images~
* [x] ~Publish Azure images; #197~
* [x] ~[Publish Google Cloud images](https://cloud.google.com/compute/docs/images/managing-access-custom-images#share-images-publicly), part of #196~


DavidGOrtega:
> Get rid of clone-amis.go in favor of Packer's inbuilt multi-region support

Originally I tried with Packer but I was not able to clone, as far as I remember but build in every region thing that was impossible due to the credentials we had at that time

images/provisioner/setup.sh
```diff
@@ -30,5 +30,5 @@ if [ ! -f "$FILE" ]; then
   sudo apt update && sudo apt install -y nvidia-docker2
   sudo systemctl restart docker
 
-  sudo echo 'OK' >"$FILE"
+  echo OK | sudo tee "$FILE"
```
DavidGOrtega, line 33: was not this addressed at #205 ?

0x2b3bfa0: Yes, please notice that this is a **draft** pull request and may include temporary upstream patches. 🙃

0x2b3bfa0:
> _Originally I tried with Packer but I was not able to clone, as far as I remember but build in every region thing that was impossible due to the credentials we had at that time_

Not sure about how it was originally, but this approach builds the image once and then clones the result, exactly like in our custom program.

Issue #224, opened 2021-09-27
Name vs Id issues with instance destruction in GCP?


DanielBarnes:
https://github.com/iterative/cml/issues/754
https://github.com/iterative/cml/issues/754#issuecomment-928240276

Possibly related:
https://github.com/iterative/cml/pull/742
https://github.com/iterative/cml/pull/738
https://github.com/iterative/terraform-provider-iterative/issues/178

Shortly after Github started Clubbing the runner name env the workaround of setting the name empty resulted in successful workflows from @0x2b3bfa0 suggestion of `RUNNER_NAME= ` then it seems with the release of CML `0.7.4` the destroy stopped working. From my activity logs on GCP when the workflows were successful the instances were named with by the Id generation found here: https://github.com/iterative/terraform-provider-iterative/blob/732d0919577867d40b7ff635dd0667caf19d3cb9/iterative/utils/helpers.go#L19

and in total it sets the name if it is empty to equal the Id:
https://github.com/iterative/terraform-provider-iterative/blob/732d0919577867d40b7ff635dd0667caf19d3cb9/iterative/utils/helpers.go#L17-L25

The code for the destroy pulls the Id https://github.com/iterative/terraform-provider-iterative/blob/732d0919577867d40b7ff635dd0667caf19d3cb9/iterative/gcp/provider.go#L273-L275

But the Golang documentation has some not exactly clear wording that makes me think this should be the name and not the Id...
https://pkg.go.dev/google.golang.org/api/compute/v1#InstancesService.Delete

From looking at the Name vs ID in my GCP activity logs it seems the delete VM has only succeed when the Name and Id were the same.

DanielBarnes:
Happy to provide any logs from GCP that may be of help upon request.

PR #225, opened 2021-09-27
#224

DanielBarnes:


DanielBarnes:
If you unset the CML_RUNNER_NAME the correct cleanup behavior is seen:
```
  CML_RUNNER_NAME= cml-runner \
    --single \
    --log=debug \
    --idle-timeout=1800 \
    --token=*** \
    --cloud=gcp \
    --cloud-region=us-west \
    --cloud-type=m \
    --cloud-hdd-size=10
```
Terraform uses the generated `iterative-xxxxxxxxx` as both the name and id which I believe confirms the behavior I've described above.

PR #251, opened 2021-11-18
Use system date in seconds for development version

0x2b3bfa0:
This will eliminate the pain of removing `.terraform.lock.hcl` between `make install` and `terraform init`


0x2b3bfa0:
Included on https://github.com/iterative/terraform-provider-iterative/pull/237

PR #256, opened 2021-11-23
Skip windows arm* builds

0x2b3bfa0:
We don't need them, and builds are broken


DavidGOrtega (approved):
🚀

casperdcl (commented):
costed an arm but not a leg(end)

PR #270, opened 2021-11-23
Turn directory tree into preformatted block

0x2b3bfa0:
**Definitely, the project layout should not look similar to _this_**
> Captura de pantalla 2021-11-24 a las 0 23 41


dacbd (approved):
👍

PR #273, opened 2021-11-24
Wait for data to be synchronized before exiting

0x2b3bfa0:
Otherwise, tests can be flaky... sometimes. ❄️

### DIFF
--- a/task/task_test.go
+++ b/task/task_test.go
@@ -76,8 +76,9 @@ func TestTask(t *testing.T) {
 					Image: "ubuntu",
 					Script: `#!/bin/bash
 						cat data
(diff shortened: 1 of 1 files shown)


task/task_test.go
```diff
@@ -77,6 +77,7 @@ func TestTask(t *testing.T) {
 					Script: `#!/bin/bash
 						cat data
 						echo "$ENVIRONMENT_VARIABLE_DATA" | tee data
+						sleep 60
```
0x2b3bfa0, line 80:
```suggestion
						cat data
						echo "$ENVIRONMENT_VARIABLE_DATA" > data
						sleep 60
						cat data
```

Otherwise, it would be ineffective: the condition to terminate the machine is the presence of bith the old and the new data strings.

PR #278, opened 2021-11-24
Disable spot on tests

0x2b3bfa0:
Infrastructure tests are flaky enough without using spot instances.


0x2b3bfa0:
Not sure it's a good idea, though. I believe that I have identified the only worrying cause of flakiness, and it isn't spot; at least, not today. 🙈

PR #323, opened 2021-12-02
Expanded logic on Subnet selection

dacbd:
- https://github.com/iterative/cml/issues/814
 - https://github.com/iterative/cml/issues/795
 - https://github.com/iterative/cml/issues/459


dacbd:
@casperdcl @DavidGOrtega @0x2b3bfa0 

thought for the below?
```go
		"aws_subnet_id": $schema.Schema{ // cloud agnostic feature? subnet_id
			Type:     schema.TypeString,
			ForceNew: true,
			Optional: true,
			Default:  "",
		},
```

iterative/utils/helpers.go
```diff
@@ -23,3 +23,13 @@ func SetId(d *schema.ResourceData) {
 		}
 	}
 }
+
+func StripAvailabilityZone(region string) string {
+	lastChar := region[len(region)-1]
+	// 0x61(a) to 0x7a(z)
+	if lastChar >= 0x61 && lastChar <= 0x71 {
```
0x2b3bfa0, line 30:
```suggestion
	if lastChar >= 'a' && lastChar <= 'z' {
```

Using [`rune` literals](https://go.dev/ref/spec#Rune_literals) could be a bit clearer and reduce the need for explanatory comments.

iterative/aws/provider.go
```diff
@@ -44,7 +48,7 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	if ami == "" {
 		ami = "iterative-cml"
 	}
-
+	fmt.Println(region)
```
DavidGOrtega, line 51: skipped by accident?

iterative/aws/provider.go
```diff
@@ -188,26 +192,48 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
...
+			Values: []string{subnetId},
+		})
+	}
+	subDesc, err := svc.DescribeSubnets(ctx, subnetOptions)
 	if err != nil {
 		return decodeAWSError(region, err)
 	}
 	if len(subDesc.Subnets) == 0 {
-		return errors.New("no subnets found")
+		return errors.New("No Subnet found")
```
DavidGOrtega, line 224: to be consistent with prior 'no ... found' checks

```suggestion
		return errors.New("no Subnet found")
```

iterative/aws/provider.go
```diff
@@ -188,26 +192,48 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
...
 	}
 	var subnetID string
-	for _, subnet := range subDesc.Subnets {
-		if *subnet.AvailableIpAddressCount > 0 && *subnet.MapPublicIpOnLaunch {
-			subnetID = *subnet.SubnetId
-			break
+	// bypass with user provided ID
+	if subnetId == "" {
```
DavidGOrtega, line 228: I think that this code is not correct. Leave it as it was.
The reason is that we need the subnet to be have a publicIp. The check is still in place and a must. 
The code is returning the first subnet that have a public IP but because you are adding a filter in your Describe priorly your subnets list will be composed by one element at most.
To that element we will do the publicip check

DavidGOrtega (changes requested):

iterative/aws/provider.go
dacbd: Yes, It will be gone by the "ready for review" time (early comments are always welcome). Also, it didn't even work 😄

iterative/aws/provider.go
dacbd: I want to bypass the public IP check by using a specific subnet-id.

My "hypothetical" reason is some Org's may require certain complex requirements on public-facing instances even with secure defaults with key-based authentication via ssh which is the only port exposed.

By providing a specific subnet-id the user should verify the instance is routable:

```
[public]        [10.0.0.0/8]
[Github]        [Self-Hosted ghar]
#  <------------ waits for jobs
push event ----> gets job                   [10.0.0./8]
                 cml runner --------------> [ CML Instance]
                                            [ephermal github runner]
# <---------------------------------------- [Training]
                                            [Instance tear down]
#  <------------ waits for jobs                

```

If want to try and have a quick meeting I can try and better explain the use case for some of these features I have been adding.

dacbd:
⚠️ As it currently stands `aws_subnet_id` *most likely* requires `aws_security_group` to be set as that is how the vpc-id is chosen.

dacbd:
Looks like these (https://github.com/iterative/terraform-provider-iterative/tree/master/iterative/testdata) would need to be regenerated.

iterative/aws/provider.go
DavidGOrtega: > I want to bypass the public IP check by using a specific subnet-id.

That would work for the `machine` but not for the `runner`
We need to check the runner readiness via SSH.

DavidGOrtega:
> Looks like these (https://github.com/iterative/terraform-provider-iterative/tree/master/iterative/testdata) would need to be regenerated.

Yes, the template tests must be regenerated 🙏 

`cd iterative && go test -update` should work

iterative/aws/provider.go
dacbd: But If `cml runner` is run from within the same subnet is will work 😸 which is the intention here. I would argue that if a user supplies an exact subnet/vpc to use the onus is on them to ensure the networking allows the connection from where they are invoking `cml runner` and thus the readiness check via SSH will work.

It should totally be documented that you could shoot yourself and create an unroutable instance.

I'm happy to connect so I can explain the use case further or alternatively, I can check our NDA status and can send a teams meeting to @casperdcl to show you guys what I mean for this environment/use-case to accomplish.

dacbd: Also @DavidGOrtega @0x2b3bfa0 my choice of `subnetId` may have been poor with `subnetID` also existing. Where `subnetId` is my user provided option, an example being: `"subnet-06ee5a8e940002c2d"` while `subnetID` is the original value derived from the `svc.DescribeSubnets` call

iterative/aws/provider.go
DavidGOrtega: I know! same happened to me long time ago

iterative/aws/provider.go
DavidGOrtega: Ahhhhh! got it. As you say this would work for on-premise instances! 🙏 

> my choice of subnetId may have been poor with subnetID also existing.

Originally I was suggesting remove the else assigning subnetID with subnetId at least. Then I realised about the publicip need.

dacbd: Exactly, I guess I was avoiding the terminology "on-premise" since this is still happening on AWS but it is "on-premise" in the sense that it involves a hybrid-multi-cloud buzzword setup.

dacbd:
I was looking at other parts of the code (`resource_runner.go`), I've only really familiarized myself with the `provider` portions. In my basic testing, everything works fine. I'm not sure what the purpose of the following: https://github.com/iterative/terraform-provider-iterative/blob/550c698677b94644a80fa019fab6f62b26787543/iterative/resource_runner.go#L363
and:
https://github.com/iterative/terraform-provider-iterative/blob/550c698677b94644a80fa019fab6f62b26787543/iterative/resource_runner.go#L445-L479

Is this something that needs to be updated as well?

iterative/aws/provider.go
```diff
@@ -387,6 +417,15 @@ func awsClient(region string) (aws.Config, error) {
 	)
 }
 
+func GetAvailabilityZone(region string) string {
+	lastChar := region[len(region)-1]
+	if lastChar >= 'a' && lastChar <= 'z' {
+		return region
+	} else {
```
DavidGOrtega, line 424: `else` is not needed just return

iterative/utils/helpers.go
```diff
@@ -23,3 +23,12 @@ func SetId(d *schema.ResourceData) {
 		}
 	}
 }
+
+func StripAvailabilityZone(region string) string {
+	lastChar := region[len(region)-1]
+	if lastChar >= 'a' && lastChar <= 'z' {
+		return region[:len(region)-1]
+	} else {
```
DavidGOrtega, line 31: `else` not needed just return

DavidGOrtega:
> Is this something that needs to be updated as well?

@dacbd  
No. Thats the code that establishes the runner as a service within the instance. Regarding to the scope of this PR we are safe.

DavidGOrtega:
@dacbd this PR is solving [#459](https://github.com/iterative/cml/issues/459) having the user to setup the region as always plus the subnet id, right?

dacbd:
Pretty  much exactly as @0x2b3bfa0 mentions in that issue. I can provide a couple example tf templates I used for testing in an hour or two.

dacbd:
Here are a few cases:

The following has no changes and works as previously
```hcl
resource "iterative_machine" "runner-test" {
    cloud = "aws"
    region = "us-west-2"
    instance_type = "t1.micro"
    ssh_private = ***
}
```
-------
Adds the availability-zone filter for to select a subnet, the `b` is stripped from the `region` to not break anything
```hcl
resource "iterative_machine" "runner-test" {
    cloud = "aws"
    region = "us-west-2b"
    instance_type = "t1.micro"
    ssh_private = ***
}
```
--------
Uses the specific subnet.
```hcl
resource "iterative_machine" "runner-test" {
    cloud = "aws"
    region = "us-west-2"
    aws_subnet_id = "subnet-qwert2345"
    instance_type = "t1.micro"
    ssh_private = ***
}
```
---------
Ignores the availability-zone because is has a subnet_id to use.
```hcl
resource "iterative_machine" "runner-test" {
    cloud = "aws"
    region = "us-west-2b"
    aws_subnet_id = "subnet-qwert2345"
    instance_type = "t1.micro"
    ssh_private = ***
}
```
------
With a security group, while not "strictly" required if you subnet is not in the "default" vpc then this is how select the vpc that your subnet-id is in, otherwise you will get an error that no subnets were found.
```hcl
resource "iterative_machine" "runner-test" {
    cloud = "aws"
    region = "us-west-2"
    aws_subnet_id = "subnet-qwert2345"
    aws_security_group = "github-actions-test"
    instance_type = "t1.micro"
    ssh_private = ***
}
```

DavidGOrtega (approved):
👍 
Thanks a lot for the effort!
🙏 ❤️

casperdcl (approved):
thanks!

PR #344, opened 2022-01-03
Fix runner termination in AWS with non cloud regions

### LINKED ISSUE #343: Runner does not destroy if we set a synthetic region in AWS
If we try an AWS region the runner is able to destroy itself, however if we setup one of our synthetic regions the runner will fail with and error like:

```
Error: Failed disposing the machine: operation error EC2: DescribeInstances, exceeded maximum number of attempts, 3, https response error StatusCode: 0, RequestID: , request send failed,  : dial tcp: lookup ec2.us-we.amazonaws.com on 127.0.0.53:53: no such host\n│ \n│ \n╵\n"}
```

dacbd:
- closes #343


dacbd:
@DavidGOrtega this should take care of it, I have to set up an AWS acc to really give it whirl if you can try it out against how you were able to reproduce the original error as well?

DavidGOrtega:
I will try it

iterative/resource_runner.go
```diff
@@ -380,7 +380,7 @@ func provisionerCode(d *schema.ResourceData) (string, error) {
 				Attributes: AttributesType{
 					ID:                 d.Id(),
 					Cloud:              d.Get("cloud").(string),
-					Region:             utils.StripAvailabilityZone(d.Get("region").(string)),
+					Region:             d.Get("region").(string),
```
DavidGOrtega, line 383: this will fail if we set `us-west-1a` we wont be able to create the client. We have to use the `utils.GetRegion(d)`

iterative/utils/helpers.go
```diff
@@ -7,6 +7,29 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var (
+	SynthRegions = map[string]map[string]string{
```
DavidGOrtega, line 11: Not sure about this... I prefer to have the GetRegions implementation per provider. Or should go in the machine... if we were using the machine...

DavidGOrtega: got it

iterative/resource_runner.go
dacbd: That is how I had it originally, I assume you got an error with it? I'm slow and haven't tested it myself yet 🙈 , I was thinking it was safe / I'll swap it back.

iterative/utils/helpers.go
dacbd: I moved it here because I think there was actually a bug in the subnet-id code after you guys fixed the last one.

See previous:
```go
	region := GetRegion(d.Get("region").(string))
	availabilityZone := GetAvailabilityZone(region)
```
The get avail func would always get a stripped region and never detect an availability-zone, now it needs to first reference the synth-regions map to eliminate those as a false positive.

Since it need to be referenced in multiple places I went ahead and broke it out / consolidated that helper func across all providers.

dacbd:
after beating my personal AWS account's 2FA (had to dig up an old phone 🔒 ) I tested with the original command from discord and the instance self terminated successfully:
```
./bin/cml.js runner \
    --labels=cml-runner \
    --idle-timeout=60 \
    --reuse \
    --cloud aws \
    --cloud-type=t2.micro \
    --cloud-spot true \
    --cloud-region us-west

```

DavidGOrtega:
Thanks @dacbd. How did you manage to setup this version of the provider in the cloud instance?
The ways I have found are:
- Create your own provider in terraform's registry
- use a local provider -> setup a large idle_timeout (10mins) -> connect SSH -> make my terraform provider

Obviously the first one is easier but the registry is buggy and sometimes the provider breaks and can not be updated and cant be deleted. I have a bunch of them already

dacbd:
@DavidGOrtega you know I didn't actually put much thought into it and now that you mention is I'm not 100% sure it worked properly(using the right one), (but it kinda, since it self-terminated). I modified the cml terraform.js to use a local one and added a git clone/checkout then make && make install into a startup script...

Were you able to test it with the latest commit successfully? -- I've put more thought in testing custom cml versions #255 😉

DavidGOrtega:
@dacbd still testing it... Right now I have an issue but it should not be because of this PR  think...
With region `us-west-1` goes fine, with `us-west` it does not... it propagates the tf file with region `us-west-1` and then I have an error about roles... a very nice riddle since as I say using `us-west-1` is working always

```
No EC2 imds role found
```

dacbd:
I'll try a few more as well

iterative/utils/helpers.go
```diff
@@ -7,6 +7,29 @@ import (
...
+			"eu-north": "eu-north-1",
+			"eu-west":  "eu-west-1",
+		},
+		"gcp": {
+			"us-east":  "us-east1-c",
+			"us-west":  "us-west1-b",
+			"eu-north": "europe-north1-a",
+			"eu-west":  "europe-west1-d",
+		},
+		"az": {
```
DavidGOrtega, line 24: @dacbd cloud in `machine` and `runner` is `azure` not `az`

@0x2b3bfa0 we should have maintained the consistency 

I do not feel confident to release this without having tested every cloud. 
Please @dacbd try to fix it just only for AWS vendor without reaching a wider scope

DavidGOrtega (changes requested):

iterative/utils/helpers.go
DavidGOrtega: I mean roll back utils.Get region

0x2b3bfa0: > _@0x2b3bfa0 we should have maintained the consistency_

We do, to some extent.

https://github.com/iterative/terraform-provider-iterative/blob/01af1440a6232c23d46fdd1967bb6d452783cd1d/iterative/resource_machine.go#L200

dacbd:
@DavidGOrtega Something like this okay or do you want it more fully reverted back?

iterative/utils/helpers.go
DavidGOrtega: > > _@0x2b3bfa0 we should have maintained the consistency_
> 
> We do, to some extent.
> 
> https://github.com/iterative/terraform-provider-iterative/blob/01af1440a6232c23d46fdd1967bb6d452783cd1d/iterative/resource_machine.go#L200

We should support then `az` in the `machine` and `runner`

0x2b3bfa0: We _do_ support `az` in both the `machine` and `runner` resources. At least, before merging this pull request.

DavidGOrtega: One should be an alias of the other to avoid issues like this one

0x2b3bfa0: Do you mean replacing the value intead of accepting both?

DavidGOrtega: Yes, an alias

dacbd:
I've had good terminations with:
```
us-west
us-west-1
us-west-1c
us-west-1b
```

iterative/utils/helpers.go
```diff
@@ -7,6 +7,17 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var (
+	AWSSynthRegions = map[string]map[string]string{
```
0x2b3bfa0, line 11: Given that it's only for AWS, wouldn't it be enough to use a `map[string]string`? Else, should it be named `SynthRegions` instead?

DavidGOrtega: Indeed probably I would keep it in the AWS provider instead of utils

dacbd: https://github.com/iterative/terraform-provider-iterative/pull/344#issuecomment-1005009172

dacbd: I can roll it back more, I left it in to inflict my feelings that these region mappings should all be defined in one place.

DavidGOrtega: Probably makes sense to have it in the provider following the same pattern as the others. Also as @0x2b3bfa0 pointed out there is not need to use a matrix

DavidGOrtega (approved):
👍

DavidGOrtega:
🙏 thanks @dacbd !!! ❤️

PR #378, opened 2022-01-28
Simplify log format

0x2b3bfa0:
https://www.notion.so/TPI-0-9-feedback-99cff02fa4b44dd9815767d953616bd1?d=bbf17979200844c2b7b8a7785ca08693#6bc5b057cfc94239aef1a10cfcdbc7e4

### DIFF
--- a/task/common/machine/script.go
+++ b/task/common/machine/script.go
@@ -107,7 +107,7 @@ sudo systemctl enable tpi-task.service --now
 
 while sleep 5; do
   journalctl > "$TPI_LOG_DIRECTORY/machine-$TPI_MACHINE_IDENTITY"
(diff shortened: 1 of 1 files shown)


DavidGOrtega:
This maybe does not solves the proposal as far as I understand it

0x2b3bfa0:
Yes, it doesn't _solve_ the proposal. It's just a small step in that direction.

DavidGOrtega:
Indeed I like it to have the date.
IMHO we need to bring customised logs with backends

0x2b3bfa0:
I also like it to have the date, similarly to GitHub Actions and any other CI/CD engine logs. 🤷🏼 Still, noise versus functionality is an open debate.

DavidGOrtega:
I do not consider date noise so I do not like this PR

casperdcl (commented):
don't see much point in removing datetime tbh.

PR #381, opened 2022-02-02
Avoid taint on apply error

0x2b3bfa0:
When `terraform apply` failed, the state file still retained a copy of the resource, and subsequent runs had no effect. 🙈  

```json
{
  ···
  "resources": [
    {
      ···
      "instances": [
        {
          "status": "tainted",
          ···
        }
      ]
    }
  ]
}
```


DavidGOrtega (approved):
👍

PR #383, opened 2022-02-07
Reboot instance if gpu is not accesible

DavidGOrtega:
In GCP the startup script is running every single boot this PR:
 - checks that cml service is running to avoid setup the instance again
 - checks is gpu is needed and reboots if not accessible  
 - closes https://github.com/iterative/cml/issues/860


0x2b3bfa0 (approved):
🎉

DavidGOrtega:
I would release now @0x2b3bfa0 we have some users hunting for this

Issue #404, opened 2022-02-19
cml: deadlock when running the script for cloning amis


redouan-rhazouani:
**Description**
Running the script "cml/clone-amis.go" might result in a deadlock.
The deadlock is caused by the go-routines not freeing the WaitGroup after completion.

**To Reproduce**
Just try to run the script.

**Expected behavior**
The deadlock is resolved

**Additional context**
See https://pkg.go.dev/sync#WaitGroup on how to use WaitGroup.

redouan-rhazouani:
Fix: https://github.com/iterative/terraform-provider-iterative/pull/405

0x2b3bfa0:
Closed with #405

PR #405, opened 2022-02-19
cml: resolve deadlock when cloning amis (#404)

redouan-rhazouani:
* Each goroutine must call wg.Done() otherwise wg.Wait() forever

### DIFF
--- a/cml/clone-amis.go
+++ b/cml/clone-amis.go
@@ -69,9 +69,10 @@ func main() {
 	ami := imagesRes.Images[0]
 
 	var wg sync.WaitGroup
+	wg.Add(len(regions))
 	for _, destination := range regions {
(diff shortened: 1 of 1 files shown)


cml/clone-amis.go
```diff
@@ -69,9 +69,10 @@ func main() {
 	ami := imagesRes.Images[0]
 
 	var wg sync.WaitGroup
+	wg.Add(len(regions))
```
0x2b3bfa0, line 72: Out of curiosity, what's the advantage of running `wg.Wait(len(slice))` before the loop instead of running `wg.Wait(1)` inside the loop?

0x2b3bfa0 (approved):
🙈 Thank you very much!

PR #418, opened 2022-03-02
Create .devcontainer.json

0x2b3bfa0:
~(For codespaces)~ I'm using [GitHub Codespaces](https://github.com/features/codespaces) to make some parts of my development workflow more reliable regardless of my local technical limitations. Terraform isn't part of the default configuration, and installing it on every new boot is a nuisance; this file solves that.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,5 +1,5 @@
 {
   "name": "terraform",
   "extensions": ["hashicorp.terraform"],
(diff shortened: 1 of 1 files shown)


DavidGOrtega (changes requested):
Could you please elaborate a bit more? Why do we need codespaces?

DavidGOrtega (approved):
👍

PR #420, opened 2022-03-03
Empty buckets during Delete()

### LINKED ISSUE #388: `task` status is confusing and not working properly
We have some race conditions aside not very clear definition in the status field that make insecure to use it to check if a task is done (despite its resources are not yet being disposed).

This is specially a bad issue for CML

0x2b3bfa0:
Would fix #388 `az` consistency issue and adds debug output to diagnose further status inconsistencies. Additionally, simplifies all the bucket emptying logic by using `rclone` instead of provider–specific code.

Side note: these changes are also a small step towards #299.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "name": "terraform",
+  "extensions": ["hashicorp.terraform"],
+  "postCreateCommand": "/bin/bash -c 'curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add - && sudo apt-add-repository \"deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main\" && sudo apt-get update && sudo apt-get install terraform'"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -41,4 +41,4 @@ terraform
 # Ignore the Go vendor directory, superseded by proxy.golang.org and go.{mod,sum}
 vendor
 
-./terraform-provider-iterative
+terraform-provider-iterative
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.8.0
 	github.com/rclone/rclone v1.57.0
 	github.com/sebdah/goldie/v2 v2.5.3
+	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
 	golang.org/x/oauth2 v0.0.0-20210819190943-2bc19b11175f
--- a/iterative/resource_task.go
+++ b/iterative/resource_task.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
+	"terraform-provider-iterative/iterative/utils"
 	"terraform-provider-iterative/task"
 	"terraform-provider-iterative/task/common"
 )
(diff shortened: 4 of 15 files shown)


DavidGOrtega:
@0x2b3bfa0 unfortunately all the issues that I had was during refresh

0x2b3bfa0:
Just to make sure we're aware of the same issue: can you please try out this branch and see if there is any status inconsistency that isn't part of the cloud expected behavior?

DavidGOrtega:
> Just to make sure we're aware of the same issue: can you please try out this branch and see if there is any status inconsistency that isn't part of the cloud expected behavior?

Sorry I do not understand. This PR might fix the status issues presented on #388, correct? However looking at the code it seems that this is happening when destroying the resource?

I have made that note because of course I can try (indeed I did) this PR but reproducing the error is not easy even in the original branch.

DavidGOrtega:
Just to be sure:
Is this PR is fixing #388 even on `terraform refresh`

DavidGOrtega:
##  Case: Task is running while the status says the opposite ❌ 
```
"logs": [
                "-- Logs begin at Mon 2022-03-07 06:42:57 UTC, end at Mon 2022-03-07 06:43:37 UTC....
              ],
"status": {
                "running": 0
              },
```

##  Case: Task is not running while the status says the opposite ⌛

DavidGOrtega:
I can confirm that on refresh the status still behaves bad @0x2b3bfa0

0x2b3bfa0:
### Task is running while the status says the opposite

This pull request does not fix this issue. It only introduces new `TF_LOG=DEBUG` log messages to debug the only part of the code that has the ability of producing a similar issue.

If/once you collect those, we can find out if it's an inconsistency on the cloud API or a missing check in our code.

0x2b3bfa0:
# Task is not running while the status says the opposite

I believed that this pull request fixed this particular issue on Azure, and all my tests were successful.

There is no (obvious) way this condition can happen without any machine running in the Scale Set nor any object in the storage bucket.

DavidGOrtega:
@0x2b3bfa0 as I also noticed fixed case b but not a (or at least in the runs I had) 
That do not closes #388 completely. Just only a part.
Does it sounds good?

0x2b3bfa0:
> _That do not closes https://github.com/iterative/terraform-provider-iterative/issues/388 completely. Just only a part. Does it sounds good?_

Yes. Debugging the [first case above](https://github.com/iterative/terraform-provider-iterative/pull/420#issuecomment-1060249099) requires, [as already stated](https://github.com/iterative/terraform-provider-iterative/pull/420#issuecomment-1060380305) merging the changes on this pull request and collecting `DEBUG` logs on runs where it can be reproduced.

0x2b3bfa0:
(Either that, or keep me trying to figure out how to reproduce it) 😅

DavidGOrtega:
This branch takes lots of time to destroy the task. Does it sounds right @0x2b3bfa0 ? Or is just coincidental? 

```
iterative_task.tpi-examples-basic: Still destroying... [id=tpi-nztgfbtfeibl-417qafg6-4ornkmtj, 6m49s elapsed]
iterative_task.tpi-examples-basic: Still destroying... [id=tpi-nztgfbtfeibl-417qafg6-4ornkmtj, 6m59s elapsed]
iterative_task.tpi-examples-basic: Still destroying... [id=tpi-nztgfbtfeibl-417qafg6-4ornkmtj, 7m9s elapsed]
iterative_task.tpi-examples-basic: Still destroying... [id=tpi-nztgfbtfeibl-417qafg6-4ornkmtj, 7m19s elapsed]
```

0x2b3bfa0:
> _This branch takes lots of time to destroy the task. Does it sounds right @0x2b3bfa0 ? Or is just coincidental?_

🤔 Precisely, this branch introduces some changes to the destroy mechanism. Now it explicitly deletes every blob on Azure Blob Storage before deleting the container.

This is expected to take seconds rather than minutes, though.

DavidGOrtega:
> This is expected to take seconds rather than minutes, though. 🙈

Weird. For me takes close to 10min almost constantly. I have to sync 20 files of 1KB each, but aside that...

0x2b3bfa0:
🥶 Only on Azure?

DavidGOrtega:
@0x2b3bfa0 K8 test is failing

0x2b3bfa0:
❄️ Flaky, not related to this pull request. <kbd>willfix</kbd>

0x2b3bfa0:
@DavidGOrtega, please always run `terraform` with `DEBUG` level logs, and post them if you hit again the _`running=0` but actually running_ condition.

0x2b3bfa0:
@DavidGOrtega: all green in 3, 2, 1, ... 🚀

PR #423, opened 2022-03-03
Logrus logs

### LINKED ISSUE #392: UX: improve log formatting & info (on refresh & show)
Log what we store in logs and events within the terraform status using the terraform CLI during refresh using [logrus](https://github.com/Sirupsen/logrus). This way would be easier to visualise the logs within the Terraform CLI.

This also applies to current existing logs in creation and destroy.

According to @dmpetrov feedback ideally:

- [ ] Instance lifecycle
     - Creation with info about price
     - Termination
- [ ] Data syncing
- [ ] Script
     - [ ] Setup environment
     - [ ] Script logs
     - [ ] Status
       - Started/Terminated
       - Success/Failed

DavidGOrtega:
The purpose is to offer an insight digest report on refresh of Task status and script logs.
 - Logrus' log level is aligned with terraform. Terraform is reading the `[LOG_LEVEL]` from the given log. Our formatter parses the desired logrus level from the call and sets the logs initial level. I.E.
    - logrus.Debug("here") => [DEBUG] TPI here
    - logrus.Info("here") => [INFO] TPI here
 - Colors and eye candy is used to make the less effort to read our logs.
 - We add a small refresh report creating a tpiFormatter that receives the `d  schemaData` and uses it to signal the logs.

### Usage
`export TF_LOG_PROVIDER=INFO` or `export TF_LOG_PROVIDER=DEBUG` prior to use terraform
### Closes
 - closes #392


DavidGOrtega:
@0x2b3bfa0 Update to your comment. Your original logs now use Logrus

image

image

DavidGOrtega:
I wonder why your logs are repeated three times. Is the refresh operation running it several times?

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+func (f *basicFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+	levelText := strings.ToUpper(entry.Level.String())
+	levelColor := colors[levelText]
+	tpl := "[%s] 🚀\x1b[%dmTPI\x1b[0m %s\n"
+	return []byte(fmt.Sprintf(tpl, levelText, levelColor, entry.Message)), nil
+}
+
+func init() {
+	colors["DEBUG"] = 36
+	colors["INFO"] = 36
```
casperdcl, line 25: probably should be different colours?

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+
+var baseTimestamp = time.Now()
+var colors = make(map[string]int)
+
+type basicFormatter struct{}
+
+func (f *basicFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+	levelText := strings.ToUpper(entry.Level.String())
+	levelColor := colors[levelText]
+	tpl := "[%s] 🚀\x1b[%dmTPI\x1b[0m %s\n"
```
casperdcl, line 19: :nail_care: would prefer

```suggestion
	tpl := "%s:🚀\x1b[%dmTPI\x1b[0m:%s\n"
```

since it leaves more room for the message.

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+	return []byte(fmt.Sprintf(tpl, levelText, levelColor, entry.Message)), nil
+}
+
+func init() {
+	colors["DEBUG"] = 36
+	colors["INFO"] = 36
+	colors["WARN"] = 33
+	colors["ERROR"] = 31
+	colors["FATAL"] = 31
+	colors["purple"] = 35
```
casperdcl, line 29: purple? why?

If it's about iterative colours should we add cyan & orange too?

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+		cloud := d.Get("cloud").(string)
+		machine := d.Get("machine").(string)
+		region := d.Get("region").(string)
+		spot := d.Get("spot").(float64)
+
+		spottext := ""
+		if spot > 0 {
+			spottext = fmt.Sprintf("(Spot %f/h)", spot)
+		}
+		message = fmt.Sprintf("🚀 %s %s%s at %s", cloud, machine, spottext, region)
```
casperdcl, line 58: The :rocket: is already in the template, right?

```suggestion
		message = fmt.Sprintf("%s %s%s in %s", cloud, machine, spottext, region)
```

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if message == "status" {
+		status := d.Get("status").(map[string]interface{})
+
+		running := "not yet started"
```
casperdcl, line 64:
```suggestion
		running := "starting 🟣"
```

or

```suggestion
		running := "queued 🟣"
```

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+		if status["running"] != nil {
+			running = "is terminated"
```
casperdcl, line 66:
```suggestion
			running = "terminated 🔵"
```

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+			if status["running"].(int) == 1 {
+				running = "is running 🟡"
```
casperdcl, line 68:
```suggestion
				running = "running 🟡"
```

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+			success = "without any output"
+			if status["succeeded"] != nil && status["succeeded"].(int) == 1 {
+				success = "succesfully 🟢"
+			}
+			if status["failed"] != nil && status["failed"].(int) == 1 {
+				success = "with errors 🔴"
+			}
+		}
+
+		message = fmt.Sprintf("Task %s %s", running, success)
```
casperdcl, line 83:
```suggestion
		if running == "terminated 🔵" {
			if status["succeeded"] != nil && status["succeeded"].(int) == 1 {
				running += " successfully 🟢"
			}
			if status["failed"] != nil && status["failed"].(int) == 1 {
				running += " with errors 🔴"
			}
		}

		message = fmt.Sprintf("Task %s", running)
```

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if message == "logs" {
+		logs := d.Get("logs").([]interface{})
+		taskLogs := "No logs"
+		if len(logs) > 0 {
+			taskLogs = strings.Replace(logs[0].(string), "\n", fmt.Sprintf("\n[%s] ", levelText), -1)
```
casperdcl, line 90:
```suggestion
			taskLogs = strings.Replace(logs[0].(string), "\n", fmt.Sprintf("\n%s:", levelText), -1)
```

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		message = fmt.Sprintf("Task logs:\x1b[%dm%s\x1b[0m", colors["purple"], taskLogs)
```
casperdcl, line 93: Purple may not be great on default Ubuntu terminals... Maybe `cyan` or `orange` instead?

iterative/utils/logger.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	tpl := "[%s] \x1b[%dm🚀TPI %s\x1b[0m %s '\n"
```
casperdcl, line 96: for consistency with previous template:

```suggestion
	tpl := "%s:🚀\x1b[%dmTPI\x1b[0m:%s:%s\n"
```

task/aws/task.go
```diff
@@ -92,139 +93,139 @@ type Task struct {
 }
 
 func (t *Task) Create(ctx context.Context) error {
-	log.Println("[INFO] Creating DefaultVPC...")
+	logrus.Debug("Creating DefaultVPC...")
```
casperdcl, line 96: are these deliberately changed from `.Info` to `.Debug`?

casperdcl (changes requested):
:rocket:

0x2b3bfa0:
> _I wonder why your logs are repeated three times. Is the refresh operation running it several times?_

Yes, it's running several times and can be optimized to run only once. See #290 for context.

iterative/utils/logger.go
DavidGOrtega: Definitely

iterative/utils/logger.go
DavidGOrtega: Let me give it a shot

iterative/utils/logger.go
0x2b3bfa0: Queued

task/aws/task.go
DavidGOrtega: Yes. I do not think that they should not be Info level. However I might be totally wrong

iterative/utils/logger.go
DavidGOrtega: This is needing to align internal TF logs with the logrus level

iterative/utils/logger.go
DavidGOrtega: It needs the spaces for readability and to align tf logs with logrus

DavidGOrtega:
please review 🙏

DavidGOrtega:
@0x2b3bfa0 might be old logging making the tests to fail?

task/aws/task.go
0x2b3bfa0: I believe that they are useful as part of the `INFO` logging level, as they provide valuable insights on the progress of creation and deletion operations.

Still, update (refresh) operations [run several times](https://github.com/iterative/terraform-provider-iterative/pull/423#issuecomment-1058574085) and add a bit of extra noise. Is this the reason why you prefer logging all these operations with the `DEBUG` level? 🤔

iterative/utils/logger.go
0x2b3bfa0: Sounds [chaotic neutral versus lawful evil](https://xkcd.com/2408/). 🤔 Applying this suggestion would make some `logrus` log lines look different to the standard library `log` ones.

```diff
- 2022-03-03T17:22:29.743Z [INFO]  provider.terraform-provider-iterative: 2022/03/03 17:22:29 [INFO] Reading ...
+ 2022-03-03T17:22:29.743Z [INFO]  provider.terraform-provider-iterative: 2022/03/03 17:22:29 INFO Reading ...
```

Still, this drawback doesn't seem to be relevant anymore after the changes introduced with cb5d9ec35be30e4d19f38fa23ee72b4f2af3db7b. Now all our user–facing logs use `logrus` and we can choose what format to apply.

0x2b3bfa0: I'd consider removing the brackets and sparing another two bytes from the required terminal width. 🤷🏼‍♂️ Still, it's probably bikeshedding at its finest.

iterative/utils/logger.go
0x2b3bfa0: Same as https://github.com/iterative/terraform-provider-iterative/pull/423#discussion_r819149028 it seems.

0x2b3bfa0:
It was just that your branch lacked some updates. Still, tests are a bit flaky and need some improvements.

0x2b3bfa0 (approved):
Looks good to me as a proof of concept. 🚀 Still, it has some unavoidable rough edges that might prevent us from providing an optimal user experience:


* Enabling this feature requires users to explicitly set the [`TF_LOG_PROVIDER`](https://www.terraform.io/internals/debugging) environment variable to `INFO` or any other appropriate value:

   **Prefixing every `terraform refresh` command**
   ```console
   $ TF_LOG_PROVIDER=INFO terraform refresh
   ···
   $ TF_LOG_PROVIDER=INFO terraform refresh
   ```

   **Using [`export`](https://pubs.opengroup.org/onlinepubs/009604599/utilities/export.html) at the beginning of every new session**
   ```console
   $ export TF_LOG_PROVIDER=INFO
   $ terraform refresh
   ···
   $ terraform refresh
   ```

* Terraform performs as many operations in parallel as possible, and declaring several resources will cause their respective log lines to be interleaved without any specific order, although there are some workarounds:

   **Using the [`--target`](https://learn.hashicorp.com/tutorials/terraform/resource-targeting?in=terraform/cli
) command–line option**
   (Runs the refresh operation for a single resource)

   ```hcl
     ···
   }
   ```
   ```console
   $ TF_LOG_PROVIDER=INFO terraform refresh --target=iterative_task.example
   ```

  **Using the [`--parallelism`](https://www.terraform.io/cloud-docs/workspaces/variables#parallelism) option**, proposed by @DavidGOrtega
   (Runs the refresh operation for all the resources, serially)

   ```console
   $ TF_LOG_PROVIDER=INFO terraform refresh --parallelism=1
   ```

* All the log lines have a lengthy 72 character prefix that can't be easily disabled:

   ```
   2022-03-03T17:22:31.220Z [INFO]  provider.terraform-provider-iterative: THIS IS WHERE ACTUAL LOG MESSAGES BEGIN
   ```

* Logs exposed through these options are meant for debugging providers, not to be used during the normal operation of the `terraform` command, although some users have been requesting this capability for a long time: 
  * https://github.com/hashicorp/terraform-plugin-sdk/issues/145
  * https://github.com/hashicorp/terraform/issues/17267

---

Bikeshedding–level subjective appreciation: Command–line tools with emoji were awesome in 2015, but they look a bit kitsch in retrospective. Once we start caring about user interface, it would be great to revisit [primer.style/cli](https://primer.style/cli) and several other [resources](https://www.notion.so/iterative/Command-line-interface-7ef6fbc438f0419982e6d26a0c6de990).

0x2b3bfa0:
(This could be a good place to explain why Terraform might not be the right tool for this project, given that we're rewriting official providers, reinventing the planner, and bypassing Terraform Core to print our own logs, but that's another story and it woulnd't be useful to divert this thread)

0x2b3bfa0:
(This could be a good place to explain why Terraform might not be the right tool for this project, given that we're rewriting official providers, reinventing the planner, and bypassing Terraform Core to print our own logs, but that's another story and it wouldn't be useful to divert this thread)

task/aws/task.go
DavidGOrtega: > Still, update (refresh) operations https://github.com/iterative/terraform-provider-iterative/pull/423#issuecomment-1058574085 and add a bit of extra noise.

Exactly that. At the time we fix it would be nice to put it on INFO level but right now is not clear for the user why its repeated

DavidGOrtega:
The ability of custom logs is something that Terraform know and indeed they [are willing to fix](https://github.com/hashicorp/terraform-plugin-sdk/issues/145#issuecomment-640832328)


> Terraform performs as many operations in parallel as possible,

Having parallel logs are normal. However for the refresh "report" they should be all them together. I have found after several trial that the logs were always fine. I mean the report stayed together without having applied any parallelism constrain.
In any case I did add the tpi id always to identify. This is also nice for new loggers in place

DavidGOrtega:
> It was just that your branch lacked some updates. Still, tests are a bit flaky and need some improvements.

Still failing GCP and K8S 😞

DavidGOrtega:
> Command–line tools with emoji were awesome in 2015, but they look a bit kitsch in retrospective

icons are better than text and easy to be followed within a bunch of text. Its not about emojis it's about icons

0x2b3bfa0:
> _The ability of custom logs is something that Terraform know and indeed they https://github.com/hashicorp/terraform-plugin-sdk/issues/145#issuecomment-640832328_

I don't see any signs of enthusiasm there. Especially after https://github.com/hashicorp/terraform/issues/17267#issuecomment-1036692773. 🤔

0x2b3bfa0:
> _Having parallel logs are normal. However for the refresh "report" they should be all them together. I have found after several trial that the logs were always fine. I mean the report stayed together without having applied any parallelism constrain.
In any case I did add the tpi id always to identify._

Yes, if you print all the logs at once and the process takes a split second, it would be moderately hard to hit this condition. Still, rest assured that it will happen. 🙃 Anyway.

DavidGOrtega:
> I don't see any signs of enthusiasm there. Especially after [hashicorp/terraform#17267 (comment)](https://github.com/hashicorp/terraform/issues/17267#issuecomment-1036692773). 🤔

"...Instead, any output from providers will need to be designed with a suitable structure that it can be translated to both a terminal-based UI and to other UI contexts such as Terraform Cloud's web UI,..."

What  distil is that they are not willing to support any output but be able to specify a closed JSON format to be displayed on the UI and or the Cloud's UI. It's pretty clear. They are not telling they are not supporting it.

DavidGOrtega:
> Still, rest assured that it will happen. 🙃 Anyway.

bikeshedding?

DavidGOrtega:
> 💯 See [`gh`](https://cli.github.com/) and [`poetry`](https://python-poetry.org/) for some elegant Unicode alternatives to emoji.

In any case I imagine that you are referring to 🚀 
I see the statuses 🔴 🟡 perfectly valid as icons

0x2b3bfa0:
> _In any case I imagine that you are referring to 🚀_
> _I see the statuses 🔴 🟡 perfectly valid as icons_

Definitely bikeshedding, but if we had full control of the terminal and outputted less text, I would prefer alternatives like `\e[31;1m•\e[0m` and `\e[33;1m•\e[0m` like in `poetry` 🤷🏼‍♂️ Let's postpone this discussion until this level of finesse becomes part of our priorities.

0x2b3bfa0:
> _What distil is that they are not willing to support any output but be able to specify a closed JSON format to be displayed on the UI and or the Cloud's UI._

And that format will be oriented towards sane, officially supported use cases, not printing thousands of log lines to the console.

0x2b3bfa0:
> _What distil is that they are not willing to support any output but be able to specify a closed JSON format to be displayed on the UI and or the Cloud's UI._

And that format will be oriented towards sane, officially supported use cases, not printing thousands of log lines to the console.

casperdcl:
> The ability of custom logs is something that Terraform know and indeed they [are willing to fix](https://github.com/hashicorp/terraform-plugin-sdk/issues/145#issuecomment-640832328)

I don't think they're closing that issue any time soon.

> All the log lines have a lengthy 72 character prefix that can't be easily disabled

So they can be disabled? How?

> Command–line tools with emoji [...] look a bit kitsch

and

> icons are better than text and easy to be followed

:100: I see the value in visual queues, however the different colours are enough for this (no need to have emoji).

DavidGOrtega:
> So they can be disabled? How?

unfortunately they can't

task/aws/task.go
casperdcl: The last one in this bunch could be info, i.e.

```diff
-logrus.Debug("Done!")
+logrus.Info("Done")
```

0x2b3bfa0:
> _So they can be disabled? How?_

Either:

1. Forking Terraform Core (missing troll face emoji), or
2. Piping the output to `awk` (ditto)

task/aws/task.go
0x2b3bfa0: And parhaps replaced by `{operation} done` where `{operation}` is ´create´ or `read` or `delete`

iterative/utils/logger.go
casperdcl: The main thing is we shouldn't have datetime & level appear twice on a line, and should replace `provider.terraform-provider-iterative` with `TPI`.

0x2b3bfa0: I would also like to have a grand piano under my bed. https://github.com/iterative/terraform-provider-iterative/pull/423#issuecomment-1060592746

task/aws/task.go
DavidGOrtega: > The last one or two in this bunch could be info, i.e.

Even repeating it three times? Could not it be a bit misleading?

iterative/utils/logger.go
casperdcl: We can definitely drop the second datetime, right?

task/aws/task.go
0x2b3bfa0: Yes, it would still be a bit confusing. It would be better to fix the cause of this issue. I'll look into it later; it's relatively easy to cache reads.

iterative/utils/logger.go
0x2b3bfa0: Yes, @DavidGOrtega already did that. See https://github.com/iterative/terraform-provider-iterative/pull/423#issuecomment-1058573507. Still, machine logs will necessarily need to have a different datetime, because what matters it the time when the log was collected, and not the time when it was displayed to the user.

casperdcl (approved):
i hate everything about this :sweat_smile:

task/aws/task.go
0x2b3bfa0: Tracked on https://github.com/iterative/terraform-provider-iterative/issues/290#issue-1062608705

0x2b3bfa0: Still, I'm tempted to switch them back to `INFO` even if there are repeated reads. That information is very useful to debug some timeout issues.

PR #442, opened 2022-03-14
Encrypt AWS images

0x2b3bfa0:
Would fix https://github.com/iterative/viewer/issues/3193 CML–related errors. Still, encryption at rest is (technically) irrelevant for public images. 🙃 After migrating to NVIDIA NGC images (#384 & #410) this won't be needed anymore. 😅

### DIFF
--- a/.github/workflows/build-ami.yml
+++ b/.github/workflows/build-ami.yml
@@ -1,7 +1,8 @@
 name: build-ami
 on: 
   push:
-    tags: ['ami*']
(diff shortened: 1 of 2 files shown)


0x2b3bfa0:
> [_AMIs with encrypted volumes cannot be made public_](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sharingamis-intro.html)

PR #445, opened 2022-03-14
Fix multiple resources logs

### LINKED ISSUE #438: `task` Logs are failing for multiple resources
For some reason logs are failing when having multiple resources seems that schema.ResourceData is nil 

```sh
panic: interface conversion: interface {} is nil, not *schema.ResourceData
```

DavidGOrtega:
- closes #438


casperdcl (approved):
well it's a fairly accurate error message so "my take on prose" is probably not needed.

Python is so much more legible... `d = data.get('d', {})['ResourceData']` will auto-raise as useful a `KeyError` if needed.

PR #460, opened 2022-03-29
Destroy instance if runner does not sync

DavidGOrtega:
Destroy the instance if the runner is unavailable

 - closes https://github.com/iterative/cml/issues/926


0x2b3bfa0 (approved):
🎉

PR #477, opened 2022-03-31
Creation/deletion time warns

DavidGOrtega:
- fixes a small part of #457 

```
 @DavidGOrtega UX: apply fails ungracefully when spot pricing is too low? (might be solved by TF_LOG_PROVIDER=INFO in this list below)
 @DavidGOrtega CLI UX/docs: clearly warn about how long it can take to provision (apply)
 @DavidGOrtega CLI UX/docs: clearly warn about how long it can take to clean up (destroy)
```

Im working to fix or improve the logic of listed below in another PRs

```
@DavidGOrtega UX: apply fails ungracefully when spot pricing is too low? (might be solved by TF_LOG_PROVIDER=INFO in this list below)
 
https://github.com/iterative/terraform-provider-iterative/issues/462
```

### DIFF
--- a/iterative/resource_task.go
+++ b/iterative/resource_task.go
@@ -18,6 +18,10 @@ import (
 	"terraform-provider-iterative/task/common"
 )
 
+var (
+	logTpl = "%s can take several minutes. If needed you might consider to increase %s timeout https://registry.terraform.io/providers/iterative/iterative/latest/docs/resources/task#timeout. Please wait..."
+)
+
 func resourceTask() *schema.Resource {
 	return &schema.Resource{
 		CreateContext: resourceTaskCreate,
@@ -160,10 +164,11 @@ func resourceTask() *schema.Resource {
 
 func resourceTaskCreate(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
 	logger := utils.TpiLogger(d)
-	if d.Get("spot").(float64) >= 0 {
-		logger.Info("Creation can take several minutes. With spot even longer, please consider to increase Create timeout. Please wait...")
-	} else {
-		logger.Info("Creation can take several minutes. Please wait...")
+	logger.Info(fmt.Sprintf(logTpl, "Creation", "Creation"))
+
+	spot := d.Get("spot").(float64)
+	if spot > 0 {
+		logger.Info("Fixed spot price does not ensure the Task to be launched. We recommend you to use auto spot price (0)")
 	}
 
 	task, err := resourceTaskBuild(ctx, d, m)
@@ -254,7 +259,7 @@ func resourceTaskRead(ctx context.Context, d *schema.ResourceData, m interface{}
 
 func resourceTaskDelete(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
 	logger := utils.TpiLogger(d)
-	logger.Info("Deletion can take several minutes. Please wait...")
+	logger.Info(fmt.Sprintf(logTpl, "Creation", "Creation"))
 
 	task, err := resourceTaskBuild(ctx, d, m)
 	if err != nil {
--- a/iterative/resource_task.go
+++ b/iterative/resource_task.go
@@ -159,6 +159,13 @@ func resourceTask() *schema.Resource {
 }
 
 func resourceTaskCreate(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
+	logger := utils.TpiLogger(d)
+	if d.Get("spot").(float64) != nil {
+		logger.Info("Creation can take several minutes. With spot even longer, please consider to increase Create timeout. Please wait...")
+	} else {
+		logger.Info("Creation can take several minutes. Please wait...")
+	}
+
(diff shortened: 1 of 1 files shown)


0x2b3bfa0:
Creation/deletion times shouldn't be different for spot and non–spot use cases. Are you sure that this is the cause of the timeout? 🤔 

Does this happen consistently with `spot >= 0` (automatic or fixed price) but not with `spot = -1` (disabled) ❓

DavidGOrtega:
> Creation/deletion times shouldn't be different for spot and non–spot use cases. Are you sure that this is the cause of the timeout? 🤔

Thats we have experienced. even with the feedback.
Im trying to ensure it that it's not a casuality.

iterative/resource_task.go
```diff
@@ -159,6 +163,14 @@ func resourceTask() *schema.Resource {
 }
 
 func resourceTaskCreate(ctx context.Context, d *schema.ResourceData, m interface{}) (diags diag.Diagnostics) {
+	logger := utils.TpiLogger(d)
+	logger.Info(fmt.Sprintf(logTpl, "Creation", "Creation"))
+
+	spot := d.Get("spot").(float64)
+	if spot > 0 {
+		logger.Info("Fixed spot price does not ensure the Task to be launched. It's recommended to use auto spot price (0)")
```
casperdcl, line 171:
```suggestion
		logger.Warn(fmt.Sprintf("Setting a maximum price `spot=%f` USD/h. Consider using auto-pricing (`spot=0`) instead.", spot))
```

iterative/resource_task.go
```diff
@@ -159,6 +163,14 @@ func resourceTask() *schema.Resource {
...
+	}
```
casperdcl, line 172:
```suggestion
		if spot < 0.05 {
			logger.Warn("Maximum spot price likely too low. Try at least `spot=0.05` USD/h or auto (`spot=0`).")
		}
	}
```

casperdcl (changes requested):

iterative/resource_task.go
DavidGOrtega: How do we know 0.05 is effective? Probably you wont never have a GPU under $1/h and we should warn about that possibility. Thats why  tried to warn the issue of setting a fixed price vs auto-bid

0x2b3bfa0: 💯 Setting arbitrary thresholds is rather useless. Warnings should be printed for either `spot > 0` or `spot < cloud.CurrentSpotBiddingPrice()` which isn't implemented nor particularly trivial to implement.

0x2b3bfa0: **TL;DR** Let's keep `spot > 0` as the warning condition.

DavidGOrtega: lets see what @casperdcl thinks

casperdcl: `spot > 0` is already warned about.
There should be an additional warning if spot is also `< something_ridiculous`.

I think `0.02` is a reasonable bare minimum based on `t2.micro`.

We could do `gpu ? 1 : 0.02` or even:

type | minimum recommended price
---|---
m | 0.45
l | 1.75
xl | 3.60
m+k80 | 1.10
l+k80 | 8.70
xl+k80 | 17.30
m+v100 | 1.80
l+v100 | 14.70
xl+v100 | 29.40

0x2b3bfa0: > _There should be an additional warning if spot is also `< something_ridiculous`_

Spot bidding prices are different across providers, and fluctuate considerably. Hardcoding what is and what isn't “ridiculous” for every instance offered by every cloud providers doesn't sound like a practical improvement.

Captura de pantalla 2022-04-01 a las 22 37 13

0x2b3bfa0: I'd rather rely on the `spot > 0` message and warn users: if they set a fixed price, they must know very well what they're doing.

iterative/resource_task.go
```diff
@@ -18,6 +18,10 @@ import (
 	"terraform-provider-iterative/task/common"
 )
 
+var (
+	logTpl = "%s may take several minutes (consider increasing `timeout` https://registry.terraform.io/providers/iterative/iterative/latest/docs/resources/task#timeout). Please wait."
```
0x2b3bfa0, line 22: Asking myself: if we suggest users to increase timeouts indiscriminately, why not do it ourselves and alter the defaults? 🤔 

Operation times are bound to data synchronization, though. It may be better to warn users so they can set the limits that better suit their data. 🙃

DavidGOrtega: > if we suggest users to increase timeouts indiscriminately

I just wanted to provide a handy link of how the operation can be extended so they can actually see that they can alter it and its operation its not inevitably broken.

> Operation times are bound to data synchronization,

That true however destroy has also timeout for me just syncing two small files.

0x2b3bfa0: > _That true however destroy has also timeout for me just syncing two small files_

#486

0x2b3bfa0: > _I just wanted to provide a handy link of how the operation can be extended so they can actually see that they can alter it and its operation its not inevitably broken._

👍🏼

casperdcl (approved):

PR #493, opened 2022-04-11
optional telemetry

### LINKED ISSUE #490: telemetry
Setup anonymised optional telemetry (similar to MLEM) to help prioritise issues

DavidGOrtega:
- closes #490 
 - Logs creation and deletion according to specs. 
 - Accepts stablished ENV var `DO_NOT_TRACK` to disable it.
 - `TPI_ANALYTICS_HOST` and `TPI_ANALYTICS_KEY` to modify the Jitsu provider if desired.

### Example in jitsu
```json
{
  "_timestamp": "2022-04-26T10:20:44.895781Z",
  "action": "task/destroy",
  "api_key": "s2s.vy0gtflakci9wbwyootfqy.7ztlcr98n9l02b55kktmncg",
  "backend": "az",
  "error": "string",
  "eventn_ctx_event_id": "cc85e2de-df4c-4eff-a909-88c899345078",
  "extra": {
    "ci": "",
    "cloud": "az",
    "cloud_disk_size": 30,
    "cloud_machine": "m",
    "cloud_region": "us-west",
    "cloud_spot": 0,
    "cloud_spot_auto": true,
    "task_duration": 57,
    "task_resumed": true,
    "task_status": {
      "running": 0,
      "succeeded": 1
    },
    "terraform_version": "v1.1.9"
  },
  "group_id": "",
  "interface": "cli",
  "os_name": "ubuntu",
  "os_version": "20.04",
  "source_ip": "90.160.138.1",
  "src": "api",
  "tool_name": "tpi",
  "tool_source": "terraform",
  "tool_version": "0.0.0",
  "user_id": "e02b9c45-4afa-40da-83e5-4228c2e1cf79"
}
```


casperdcl (commented):
is it worth putting this in a separate package on the offchance we

- need it in another Go pkg, or
- release a telemetrics product?

DavidGOrtega:
It's an interesting discussion. To be honest Im not sure about this. 
I understand that we might do everything visibly for the user to be able to see that we are not bluffing her. However I can see that we are duplicating efforts writing the same in different languages when we might write a pipeline within Jutsi like the ip geolocation. 
Indeed in the case of the IP the user MUST trust that we are good fella and we are not storing the IP for real, so why not for the rest of the data...

iterative/utils/analytics.go
```diff
@@ -0,0 +1,144 @@
...
+func SendJitsuEvent(eventType string, eventName string, d *schema.ResourceData) {
+	if _, ok := os.LookupEnv("DO_NOT_TRACK"); ok {
+		logrus.Debug("Analytics: DO_NOT_TRACK enabled")
+		return
+	}
+
+	postBody, _ := json.Marshal(JitsuEventPayload(eventType, eventName, d))
+
+	host := "https://0f82a78ac1e6.ngrok.io"
+	token := "s2s.vy0gtflakci9wbwyootfqy.7ztlcr98n9l02b55kktmncg"
```
DavidGOrtega, line 137: @shcheklein  @mike0sv  @iterative/cml we should use the same Jitsu right?

iterative/utils/analytics.go
```diff
@@ -0,0 +1,144 @@
...
+func TaskDuration(logs string) float64 {
+	regex := regexp.MustCompile(`\w{3} \d{2} \d{2}:\d{2}:\d{2}`)
+	matches := regex.FindAllString(logs, -1)
+	taskDuration := 0.0
+
+	if len(matches) > 1 {
+		layout := "Mar 07 06:43:27"
+		t1, _ := time.Parse(layout, matches[len(matches)-1])
+		t0, _ := time.Parse(layout, matches[0])
+		taskDuration = t1.Sub(t0).Seconds()
```
DavidGOrtega, line 38: @0x2b3bfa0 currently task has a date format that is not ideal to parse it.

iterative/utils/analytics.go
```diff
@@ -0,0 +1,144 @@
...
+	hostStat, _ := host.Info()
+	id := hostStat.HostID
+
+	if IsCI() {
+		id = fmt.Sprintf("%s%s%s",
+			os.Getenv("GITHUB_REPOSITORY_OWNER"),
+			os.Getenv("CI_PROJECT_ROOT_NAMESPACE"),
+			os.Getenv("BITBUCKET_WORKSPACE"),
+		)
+	}
```
DavidGOrtega, line 84: According to the specs we should use multiple vars to even gather the userid. I think that with the organization it suffices. Not sure why we want to dig into specific users within the organization

iterative/utils/analytics.go
```diff
@@ -0,0 +1,144 @@
...
+	} else {
+		logrus.Error("Analytics: Failed extracting terraform version")
+	}
+
+	return ""
+}
+
+func JitsuUserId() string {
+	hostStat, _ := host.Info()
+	id := hostStat.HostID
```
DavidGOrtega, line 76: A file based mechanism alternative could be anonymise the computer ID

@shcheklein @mike0sv https://github.com/orgs/iterative/teams/cml

iterative/utils/analytics.go
0x2b3bfa0: The other variables are needed for [`user_id`](https://www.notion.so/iterative/Specification-1a0197b8e34d43a89777a4c9a00ad1fe#7f4e4247c0ef4331a9243cbdb3792b18) but what you have is `group_id` more or less. 🤔 Please take another look to [this](https://www.notion.so/iterative/Specification-1a0197b8e34d43a89777a4c9a00ad1fe#c4a1c6a1e0b94d0d972e5e728fcc02d8).

iterative/utils/analytics.go
0x2b3bfa0: If you want this feature, you may prefer to write an object to object storage with the start date and the end date.

DavidGOrtega: Yes, I find total time average a very useful insight

0x2b3bfa0: Low–hanging fruit: if you want to know whether people uses spot instance recovery, you can always `if spot: len([]logs)` because every instance in the chain creates a log.

iterative/utils/analytics.go
DavidGOrtega: Thats what I say 

`GITHUB_REPOSITORY_OWNER ` is the GH organization or the user
`CI_PROJECT_ROOT_NAMESPACE` mean group/subgroup/subgroup if any
`BITBUCKET_WORKSPACE` is like GL's one

DavidGOrtega: What do we understand for user_id?

0x2b3bfa0: `user_id` is `anonymize('DavidGOrtega <g···d@gmail.com>')`
`group_id` is `anonymize('https://github.com/iterative')`

DavidGOrtega: thats what I say. 
I should only need to have `iterative`, I do not care about `DavidGOrtega`

0x2b3bfa0: I find both useful, because we can tell apart the number of contributors (e.g. size of an organization)

0x2b3bfa0: This sounds like a good topic to comment [on Notion](https://www.notion.so/iterative/Specification-1a0197b8e34d43a89777a4c9a00ad1fe#7f4e4247c0ef4331a9243cbdb3792b18)

DavidGOrtega: > because we can tell apart the number of contributors (e.g. size of an organization)

How er can tell? The id is anonymised

0x2b3bfa0: Count unique `user_id` values for a given `group_id` ❓

iterative/utils/analytics.go
casperdcl:
```suggestion
	host := "https://iterative-telemetry.herokuapp.com/api/v1/s2s/event"
	token := "s2s.jtyjusrpsww4k9b76rrjri.bl62fbzrb7nd9n6vn5bpqt"
```

.github/workflows/release.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
-          args: release --rm-dist
+          args: release --rm-dist -ldflags="-X 'terraform-provider-iterative/iterative/utils.Version=${GITHUB_REF##refs/tags/v}'"
```
DavidGOrtega, line 20: @0x2b3bfa0 this should set the Version var in the utils library to be  used in `tool_version`

0x2b3bfa0: Yup

DavidGOrtega: I think its cleaner `${{ github.ref_name}}` vs `${GITHUB_REF##refs/tags/v}`

DavidGOrtega:
Please review [updated PR description](https://github.com/iterative/terraform-provider-iterative/pull/493#issue-1200439755) 👆

iterative/resource_task.go
```diff
@@ -173,6 +173,7 @@ func resourceTaskCreate(ctx context.Context, d *schema.ResourceData, m interface
 
 	task, err := resourceTaskBuild(ctx, d, m)
 	if err != nil {
+		utils.SendJitsuEvent("task_create", err, d)
```
dacbd, line 176: Correct me if I'm wrong, but aren't these all blocking operations, I recall reading that they should not?
```suggestion
		go utils.SendJitsuEvent("task_create", err, d)
```

iterative/resource_task.go
```diff
@@ -263,14 +266,17 @@ func resourceTaskDelete(ctx context.Context, d *schema.ResourceData, m interface
 
 	task, err := resourceTaskBuild(ctx, d, m)
 	if err != nil {
+		utils.SendJitsuEvent("task_delete", err, d)
```
dacbd, line 269:
```suggestion
		go utils.SendJitsuEvent("task_delete", err, d)
```

iterative/resource_task.go
```diff
@@ -263,14 +266,17 @@ func resourceTaskDelete(ctx context.Context, d *schema.ResourceData, m interface
...
 		return diagnostic(diags, err, diag.Error)
 	}
 
 	if err := task.Delete(ctx); err != nil {
+		utils.SendJitsuEvent("task_delete", err, d)
```
dacbd, line 274:
```suggestion
		go utils.SendJitsuEvent("task_delete", err, d)
```

iterative/resource_task.go
```diff
@@ -263,14 +266,17 @@ func resourceTaskDelete(ctx context.Context, d *schema.ResourceData, m interface
...
 		return diagnostic(diags, err, diag.Error)
 	}
 
-	return
+	utils.SendJitsuEvent("task_delete", err, d)
```
dacbd, line 278:
```suggestion
	go utils.SendJitsuEvent("task_delete", err, d)
```

dacbd (changes requested):
Am I still ✈️ 😴 ?

casperdcl (changes requested):
- action="task_delete": should be `task/destroy`, `task/apply`, etc.
- extra.ci=false: would use blank string (for consistency with presumably `"github/gitlab/bitbucket"`)
- extra.cloud*: could any of this be sensitive?
- os_name=null: can we fill this in?
- source_ip="90.160.138.1": this should not be passed/saved at all?
- src="api": remove this key
- tool_source="terraform": what are the alternatives/options?
- tool_version="v0.0.0": unlikely 0.0.0 surely?

DavidGOrtega:
> * action="task_delete": should be `task/destroy`, `task/apply`, etc.
OK.

> * extra.ci=false: would use blank string (for consistency with presumably `"github/gitlab/bitbucket"`)
I will check again the specs. I did not understand that when I read it.

> * extra.cloud*: could any of this be sensitive?
I think none of them are sensitive. None of them can give us specific details of our anonymised users. 

> * os_name=null: can we fill this in?
Weird! this is a bug definitely but i have seen it previously populated. Might be related to run it inside a container? I will check

> * source_ip="90.160.138.1": this should not be passed/saved at all?
Controlled by Jitsu. It's the most worrying aspect for me. Im not including the IP by any meaning, indeed Im calling the endpoint with the ip hidden flag, but its jitsu the one that is handling this. Probably in our real instance (im using my dev this is not happening)

> * src="api": remove this key
Controlled by Jitsu. Jitsu sends the api key.

> * tool_source="terraform": what are the alternatives/options?
I was not sure about this field. But we do not yet have other options. It's right now filled to conform the given multi tool specs

> * tool_version="v0.0.0": unlikely 0.0.0 surely?
Version will be fixed in every release via  -ldflags (please read the note I have discussed with @0x2b3bfa0)

@casperdcl

iterative/resource_task.go
DavidGOrtega: I do not think that its a huge deal since we are just sending data like calling Cloud APIs, I would not say that calling it once per creation/destroy is a lot. But I will check it

.github/workflows/release.yml
0x2b3bfa0: #537

iterative/resource_task.go
DavidGOrtega: Why?

0x2b3bfa0: Fallacy number 1: _the network is reliable_ :wink:

0x2b3bfa0:
> _I was not sure about this field. But we do not yet have other options._

It's fine, to my mind. 💯

iterative/resource_task.go
DavidGOrtega: > Fallacy number 1: the network is reliable

Wont fail everything including the Vendors?

> If our telemetry server goes down, we don't want our product to fail

It works that way right now. I tested that to be sure that if the telemetry is not up we are not breaking the code

0x2b3bfa0: > _Wont fail everything including the Vendors?_

If our telemetry server goes down, cloud providers may still work.

0x2b3bfa0: > _If the telemetry is not up we are not breaking the code_

Connections can hang (again, network is unreliably), and `net/http` doesn't have any timeout by default. ✈️

DavidGOrtega:
> action="task_delete": should be task/destroy, task/apply, etc.
OK.

I followed code structure: 
 - `resourceTaskDelete`
 - `resourceTaskCreate`
 - 
Sounded to me me better than terraformic

0x2b3bfa0:
@casperdcl, I wonder if we should use the “standard” `create/delete` instead of `apply/destroy` so it stays the same for any `task` backend we add in the future.

0x2b3bfa0 (changes requested):
Please swap `idgen` with code from [this page](https://www.notion.so/iterative/User-UUID-generation-0e054a05e717409e82386f17757a2f2a).

DavidGOrtega:
review 🙏

iterative/utils/analytics.go
```diff
@@ -0,0 +1,209 @@
...
+		"group_id":     GroupId(),
+		"action":       action,
+		"interface":    "cli",
+		"tool_name":    "tpi",
+		"tool_source":  "terraform",
+		"tool_version": Version,
+		"os_name":      systemInfo["plattform"],
+		"os_version":   systemInfo["platform_version"],
+		"backend":      extra["cloud"],
+		"error":        err,
```
0x2b3bfa0, line 218: Potentially sensitive?

DavidGOrtega: error? 
maybe... I think that we hide all the sensitive data within terraform but not sure at what level it is

DavidGOrtega: error is defined in the specs

0x2b3bfa0: For example in AWS this may contain autorization errors in plain text (i.e. after STS decoding them)

DavidGOrtega: Should we avoid them then?

0x2b3bfa0: > _error is defined in the specs_

Yes, and explicitly _“must be stripped of sensitive info e.g. `str(type(exception))`”_

0x2b3bfa0: > _Should we avoid them then?_

No, just sanitize them. Although Go error management model makes this a bit difficult, to put it mildly.

DavidGOrtega: so replace it with `string(type(error)))`?

DavidGOrtega: > No, just sanitize them. Although Go error management model makes this a bit difficult, to put it mildly.

Can you please elaborate what do you have in mind?

0x2b3bfa0: `reflect.TypeOf(err).Name()` would be something like `RequestError` for SDK errors (depends on the provider)

0x2b3bfa0: If we want to provide more insightful information, we can either send the whole logs (handle with care) or try to extract anonymized information from error messages; for example, match some strings.

iterative/resource_task.go
```diff
@@ -263,14 +266,17 @@ func resourceTaskDelete(ctx context.Context, d *schema.ResourceData, m interface
...
 		return diagnostic(diags, err, diag.Error)
 	}
 
 	if err := task.Delete(ctx); err != nil {
+		go utils.SendJitsuEvent("task/destroy", err, d)
 		return diagnostic(diags, err, diag.Error)
 	}
 
-	return
+	go utils.SendJitsuEvent("task/destroy", err, d)
```
DavidGOrtega, line 278: @0x2b3bfa0 @dacbd  with `go` we are not hitting jitsu. No sure if its because we finish prior to the call... When I remove it I can see the logs in my dev server

iterative/resource_task.go
```diff
@@ -263,14 +266,17 @@ func resourceTaskDelete(ctx context.Context, d *schema.ResourceData, m interface
...
+	return diags
```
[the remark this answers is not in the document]
0x2b3bfa0: 🙅🏼 It's a [named return](https://go.dev/tour/basics/7), please leave it as–is

iterative/resource_task.go
0x2b3bfa0: Yes, @DavidGOrtega, it's a bit more involved that adding `go` and forgetting about it. I'll open a cascading pull request after an upcoming meeting.

iterative/resource_task.go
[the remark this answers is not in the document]
DavidGOrtega: Ok but we are returning the diags in others... should we be consistant?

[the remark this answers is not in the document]
0x2b3bfa0: Yes, it should be named returns everywhere. 🙈

iterative/resource_task.go
DavidGOrtega: Nah, let me try it

0x2b3bfa0: 👌🏼

iterative/resource_task.go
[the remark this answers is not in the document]
DavidGOrtega: Ok I will revert these and open a follow up `technical debt` issue

0x2b3bfa0:
> _you can drop your opinions here_

[`DROP TABLE opinions;`](https://www.youtube.com/watch?v=l7-6m2cE6JM) 😄

dacbd (commented):
LGTM, I agree with: https://github.com/iterative/terraform-provider-iterative/pull/493#pullrequestreview-952977449

Perhaps in `SendJitsuEvent` ending with:
```go
	go client.Post(url, "application/json", bytes.NewBuffer(postBody))
```
is a happy medium between an easy go and forget about it and a more complex solution?

0x2b3bfa0 (changes requested):
`s/regular expressions/goroutines/` https://github.com/iterative/terraform-provider-iterative/pull/538

<details> <summary>Relevant XKCD image...</summary>

regular_expressions

</details>

iterative/resource_task.go
0x2b3bfa0: Opened https://github.com/iterative/terraform-provider-iterative/pull/538 after https://github.com/iterative/terraform-provider-iterative/pull/493#issuecomment-1109800668

iterative/utils/analytics.go
```diff
@@ -0,0 +1,290 @@
...
+
+func JitsuEventPayload(action string, e error, extra map[string]interface{}) map[string]interface{} {
+	systemInfo := SystemInfo()
+
+	extra["ci"] = guessCI()
+	extra["terraform_version"] = TerraformVersion()
+
+	err := ""
+	if e != nil {
+		err = reflect.TypeOf(e).String()
```
0x2b3bfa0, line 214: ### Continuation of https://github.com/iterative/terraform-provider-iterative/pull/493#discussion_r858526582

If we want to provide more insightful information, we can either send the whole logs (handle with care) or try to extract anonymized information from error messages; for example, match some strings.

DavidGOrtega:
Is there any other change this PR might need?

iterative/utils/analytics.go
```diff
@@ -0,0 +1,290 @@
...
+		return ""
+	}
+
+	return id.String()
+}
+
+func UserId() string {
+	id := uuid.New().String()
+	old := appdirs.UserConfigDir("dvc/user_id", "iterative", "", false)
+	new := appdirs.UserConfigDir("telemetry", "iterative", "", false)
```
0x2b3bfa0, line 152:
```suggestion
	new := appdirs.UserConfigDir("iterative/user_id", "", "", false)
```

...or something; the specification has changed.

0x2b3bfa0: Also, @casperdcl, this file has JSON inside, right?

iterative/utils/analytics.go
```diff
@@ -0,0 +1,290 @@
...
+		"backend":      extra["cloud"],
+		"error":        err,
+		"extra":        extra,
+	}
+
+	return payload
+}
+
+func SendJitsuEvent(action string, e error, extra map[string]interface{}) {
+	for _, env := range []string{"DO_NOT_TRACK", "DVC_NO_ANALYTICS"} {
```
0x2b3bfa0, line 236: There are some [discussions](https://iterativeai.slack.com/archives/C01APS0FHDM/p1629152270013500) regarding the adopition of `DO_NOT_TRACK` we should revisit before including it.

0x2b3bfa0 (changes requested):

iterative/utils/analytics.go
casperdcl:
```suggestion
	new := appdirs.UserConfigDir("iterative/telemetry", "", "", false)
```

casperdcl: > this file has JSON inside, right?

no, just the id I think.

0x2b3bfa0: But the DVC file contains `{"user_id": "00···0"}` and we're just copying it over the new one. 🤔

0x2b3bfa0: Should we parse the JSON of the old format and write only the raw value to the new format?

casperdcl (changes requested):
missing `task/refresh`?

0x2b3bfa0:
Missing `task/refresh` is intentional, I believe. To avoid overloading our backend with too many records. 🙃

iterative/resource_task.go
```diff
@@ -188,6 +188,7 @@ func resourceTaskCreate(ctx context.Context, d *schema.ResourceData, m interface
 		}
 	}
 
+	utils.SendJitsuEvent("task/apply", err, utils.ResourceData(d))
 	return
 }
```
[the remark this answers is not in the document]
DavidGOrtega: I can be a lot of noise... not sure about it.
We might end up with huge logs and a expensive database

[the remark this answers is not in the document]
0x2b3bfa0: > _I'd say ~5 calls per `terraform refresh` is acceptable, no?_

Events are being emitted from the Terraform provider code, so it's more like ~1 call per `terraform refresh`

iterative/utils/analytics.go
DavidGOrtega: Can we unblock those discussions?

0x2b3bfa0: Yes, let me ping people [here](https://iterativeai.slack.com/archives/C6YHPP2TB/p1652206667050159?thread_ts=1652196801.349989&cid=C6YHPP2TB).

0x2b3bfa0: Created a thread [here](https://iterativeai.slack.com/archives/C0209RF6A5B/p1652345394951249)

iterative/utils/analytics.go
DavidGOrtega: I do not follow that @0x2b3bfa0 
DO you say that we should try to decode the JSON (if any) and then write the raw id as it is right now?

iterative/utils/analytics.go
DavidGOrtega: Maybe we can remove it for now and if we want to include in the future everything will be OK
At least we can unblock this

iterative/resource_task.go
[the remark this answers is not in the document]
DavidGOrtega: I have added the stats in read also

iterative/utils/analytics.go
0x2b3bfa0: It is a big effort and, yes, it will require regular expressions. It would involve collecting as many errors as possible (from our own runs) and then finding out what we can send without including sensitive information.

Sounds like a good follow-up pull request once I feel really bored. 😉

DavidGOrtega: so we are ok right now?

0x2b3bfa0: Yes, this should probably be a follow-up issue, but not something to be addressed immediately.

iterative/utils/analytics.go
0x2b3bfa0: 👍🏼

iterative/utils/analytics_test.go
```diff
@@ -0,0 +1,90 @@
...
+	info := SystemInfo()
+	assert.NotNil(t, info["platform"])
+	assert.NotNil(t, info["platform_version"])
+}
+
+func TestUserId(t *testing.T) {
+	old := appdirs.UserConfigDir("dvc/user_id", "iterative", "", false)
+	new := appdirs.UserConfigDir("iterative/telemetry", "", "", false)
+
+	userId := "a7140860-ace7-5e2a-be2b-e078074a06fd"
```
casperdcl, line 34: https://xkcd.com/221/

DavidGOrtega:
This is ready for review. The only doubts that I have right now is the Id.
As far as I understood the id is read from the file

For CI
anon hash from env vars
For non CI
read `appdirs.UserConfigDir("/iterative/telemetry", "",  "", false)` 
if not exists `appdirs.UserConfigDir("dvc/user_id", "iterative", "", false)` where iterative is just another level in the path only for Windows from a json `user_id`
If not then plain 36 hash id

Am I correct? Did I understood the current id extraction?

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+		"backend":      extra["cloud"],
+		"error":        err,
+		"extra":        extra,
+	}
+
+	return payload
+}
+
+func SendJitsuEvent(action string, e error, extra map[string]interface{}) {
+	for _, env := range []string{"DVC_NO_ANALYTICS"} {
```
casperdcl, line 248:
```suggestion
	for _, env := range []string{"ITERATIVE_DO_NOT_TRACK"} {
```

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+
+	go send(JitsuEventPayload(action, e, extra))
+	wg.Add(1)
+}
+
+func send(event interface{}) error {
+	defer wg.Done()
+
```
casperdcl, line 261: should check `if event.user_id == "do-not-track" { return nil }`

casperdcl (changes requested):

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+)
+
+const (
+	Timeout  = 5 * time.Second
+	Endpoint = "https://telemetry.cml.dev/api/v1/s2s/event?ip_policy=strict"
+	Token    = "s2s.jtyjusrpsww4k9b76rrjri.bl62fbzrb7nd9n6vn5bpqt"
+)
+
+var (
+	Version string = "0.0.0"
```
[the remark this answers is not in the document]
DavidGOrtega: No. This is updated automatically via gorelease

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+	hostStat, _ := host.Info()
+	return map[string]interface{}{
+		"platform":         hostStat.Platform,
+		"platform_version": hostStat.PlatformVersion,
+	}
+}
+
+func TaskDuration(logs string) float64 {
+	regex := regexp.MustCompile(`\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}`)
+	matches := regex.FindAllString(logs, -1)
```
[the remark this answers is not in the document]
DavidGOrtega: Um, I not sure. At most could be dateTimeRegex. dateTime reflects to me a var of type date?

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+		t1, _ := time.Parse(layout, matches[len(matches)-1])
+		t0, _ := time.Parse(layout, matches[0])
+		taskDuration = t1.Sub(t0).Seconds()
+	}
+
+	return taskDuration
+}
+
+func IsCI() bool {
+	if _, ok := os.LookupEnv("CI"); ok {
```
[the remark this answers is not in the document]
DavidGOrtega: But the rest of the CIs supports CI so we should just include `TF_BUILD` no?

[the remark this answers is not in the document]
DavidGOrtega: In any case we do not support any other CI than BB, GH and GL just to mention

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+		return ""
+	}
+
+	id, err := deterministic(fmt.Sprintf("%s%s%s%s%s",
+		os.Getenv("GITHUB_SERVER_URL"),
+		os.Getenv("GITHUB_REPOSITORY_OWNER"),
+		os.Getenv("CI_SERVER_URL"),
+		os.Getenv("CI_PROJECT_ROOT_NAMESPACE"),
+		os.Getenv("BITBUCKET_WORKSPACE"),
+	))
```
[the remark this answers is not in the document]
DavidGOrtega: Is equivalent? Don't need the if/else statements

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+		uid, err := deterministic(fmt.Sprintf("%s%s%s",
+			os.Getenv("GITHUB_ACTOR"),
+			os.Getenv("GITLAB_USER_ID"),
+			os.Getenv("BITBUCKET_STEP_TRIGGERER_UUID"),
+		))
+
+		if err == nil {
+			return uid.String()
+		}
+	}
```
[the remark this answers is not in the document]
DavidGOrtega: I would not like to rely on executing  git when we have variables that already give it to us

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+				defer jsonFile.Close()
+			}
+		}
+
+		os.MkdirAll(filepath.Dir(new), 0644)
+		ioutil.WriteFile(new, []byte(id), 0644)
+	} else {
+		dat, _ := ioutil.ReadFile(new)
+		id = string(dat[:])
+	}
```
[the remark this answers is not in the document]
DavidGOrtega: I do not follow here. Could you please elaborate?
As far as I know we should just retrieve the new format id that is stored raw and if not we have to open the old format file containing the id to be rescued from there and generate a new format file.
Its that correct?

iterative/utils/analytics.go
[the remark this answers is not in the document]
casperdcl: it's not equivalent in the sense that we want to use the syntax `A/B` not `AB`. The slash is important.

iterative/utils/analytics.go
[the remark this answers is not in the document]
casperdcl: just being nitpicky. Feel free to close unresolved :)

iterative/utils/analytics.go
```diff
@@ -0,0 +1,302 @@
...
+
+	id := uuid.NewSHA1(ns, []byte(hex.EncodeToString(dk)))
+	return &id, nil
+}
+
+func SystemInfo() map[string]interface{} {
+	hostStat, _ := host.Info()
+	return map[string]interface{}{
+		"platform":         hostStat.Platform,
```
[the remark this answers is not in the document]
casperdcl: k. Ideally we need just `windows|mac|linux` for consistency with other projects.

iterative/utils/analytics.go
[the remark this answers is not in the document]
casperdcl: I think at least `TF_BUILD` is needed - even though we don't explicitly test/"support" it.

iterative/utils/analytics.go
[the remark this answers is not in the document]
DavidGOrtega: Ah! Why the slash is important?
At the end we are having an anonymous uuid so the outcome is 0000-0000-0000... why does it matter to have the generation using a slash?

iterative/utils/analytics.go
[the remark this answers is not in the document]
DavidGOrtega: I have to save the old format too. Understood

iterative/utils/analytics.go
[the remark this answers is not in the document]
casperdcl: ​because the uuid is deterministically generated in precisely the same way in other libs.

iterative/utils/analytics.go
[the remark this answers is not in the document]
casperdcl: I mean if the *only* env var is `CI=true` then it's an unknown CI and we should set user to be `git log -1 --pretty=format:'%ae'`.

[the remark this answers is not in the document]
casperdcl: I mean if the *only* env var is `CI=true` then it's an unknown CI and we should set user to be `git log -1 --pretty=format:'%ae'`.

iterative/utils/analytics.go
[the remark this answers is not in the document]
0x2b3bfa0: Concatenating unsanitized values without separators may lead to identical results for different inputs; with separators it's also possible, but more difficult.

### ❌ Concatenation without separators 

$A \parallel B$ 

```console
% A="123"
% B="456"
% echo "$A$B" | sha256sum
e150a1ec81e8e93e1eae2c3a77e66ec6dbd6a3b460f89c1d08aecf422ee401a0  -
% A="12"
% B="3456"
% echo "$A$B" | sha256sum
e150a1ec81e8e93e1eae2c3a77e66ec6dbd6a3b460f89c1d08aecf422ee401a0  -
```

### ⚠️ Concatenation with separators

```console
% A="123"
% B="456/"
% echo "$A/$B" | sha256sum
b1c6968367cb68552685f8d84e281f1ecb5c16ffbac2d74e84f2477cac8ff7f1  -
% A="123/456"
% B=""
% echo "$A/$B" | sha256sum
b1c6968367cb68552685f8d84e281f1ecb5c16ffbac2d74e84f2477cac8ff7f1  -
```

$A \parallel separator \parallel B$ 

### ✅ Concatenation of fixed-length digests

```console
% A="123"
% B="456"
% echo "$(echo "$A" | sha256sum)" "$(echo "$B" | sha256sum)" | sha256sum
cd2aa21498289faa8940091e05f9c0f4a0f39d45310b0d251ee345a51628188e  -
```

$H(A) \parallel H(B)$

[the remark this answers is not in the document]
0x2b3bfa0: While this use case isn't quite critical, I'd recomment one of the last two options.

iterative/utils/analytics.go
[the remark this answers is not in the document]
0x2b3bfa0: Also known as “available in the standard library” through [`runtime.GOOS`](http://golang.org/pkg/runtime/#GOOS) 🙃

iterative/utils/analytics.go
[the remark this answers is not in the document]
DavidGOrtega: When is that happening? CI ENV vars can always include an specific VAR to that CI. Im a bit lost

iterative/utils/analytics.go
DavidGOrtega: Its handled in 

```go
func SendJitsuEvent(action string, e error, extra map[string]interface{}) {
	for _, env := range []string{"ITERATIVE_DO_NOT_TRACK"} {
		if _, ok := os.LookupEnv(env); ok {
			logrus.Debugf("analytics: %s environment variable is set; doing nothing", env)
			return
		}
	}

	go send(JitsuEventPayload(action, e, extra))
	wg.Add(1)
}
```

DavidGOrtega:
@iterative/cml I think i'm covering the id generation as specified and the other issues 🤞

iterative/utils/analytics.go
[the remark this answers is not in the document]
0x2b3bfa0:
```suggestion
    _, ciIsSet := os.LookupEnv("CI")
    _, tfBuildIsSet := os.LookupEnv("TF_BUILD")
    if ciIsSet || tfBuildIsSet {
```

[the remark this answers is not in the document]
0x2b3bfa0:
```suggestion
	if os.Getenv("CI") == "" || os.Getenv("TF_BUILD") {
```

For the record, I suggested this during the meeting. It's not functionally equivalent, but pretty close.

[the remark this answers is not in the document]
0x2b3bfa0: Admittedly, this in Python would look nicer; e.g. [`if {"CI", "TF_BUILD"} < set(os.environ):`](https://github.com/iterative/studio/blob/7f209862539612356527b50f1b83d6bb7ba8c47b/.circleci/build_and_push_docker_images.py#L91)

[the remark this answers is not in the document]
DavidGOrtega: Its not needed this check is using 
```go
func IsCI() bool {
	if _, ok := os.LookupEnv("CI"); ok {
		return true
	}

	if len(guessCI()) > 0 {
		return true
	}

	return false
}
```
And guessCI is checking `TF_BUILD`

```go
if _, ok := os.LookupEnv("TF_BUILD"); ok {
		return "azure"
	}

```
```

File: iterative/utils/analytics.go [hunk 27]
```diff
@@ -0,0 +1,351 @@
...
+		return "gitlab"
+	}
+
+	if _, ok := os.LookupEnv("BITBUCKET_WORKSPACE"); ok {
+		return "bitbucket"
+	}
+
+	if _, ok := os.LookupEnv("TF_BUILD"); ok {
+		return "azure"
+	}
```
casperdcl, line 111: in light of https://github.com/iterative/terraform-provider-iterative/pull/493/files#r906232930

```suggestion
	return len(guessCI()) > 0
}

func guessCI() string {
	if _, ok := os.LookupEnv("GITHUB_SERVER_URL"); ok {
		return "github"
	}

	if _, ok := os.LookupEnv("CI_SERVER_URL"); ok {
		return "gitlab"
	}

	if _, ok := os.LookupEnv("BITBUCKET_WORKSPACE"); ok {
		return "bitbucket"
	}

	if _, ok := os.LookupEnv("TF_BUILD"); ok {
		return "azure"
	}

	if _, ok := os.LookupEnv("CI"); ok {
		return "unknown"
	}
```

iterative/utils/analytics.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if os.IsNotExist(errorOld) {
+		os.MkdirAll(filepath.Dir(old), 0644)
+		data := map[string]interface{}{
+			"user_id": id,
+		}
+		file, _ := json.MarshalIndent(data, "", " ")
+		ioutil.WriteFile(old, file, 0644)
+	}
```
casperdcl, line 232: ideal pseudo-code for this is meant to be:

```python
if new.exists():
    uid = new.read_text().strip()
else:  # if uid.lower() != "do-not-track":
    if old.exits():
        uid = json.load(old.open())["user_id"]
    else:  # only for non-DVC packages
        # write legacy file in case legacy DVC is installed later
        old.write_text(f'{{"user_id": "{uid}"}}')
    new.write_text(uid)
```

yet instead the current implementation is:

```python
if not new.exists():
    if old.exits():
        uid = json.load(old.open())["user_id"]
    new.write_text(uid)
else:  # new exists - ignore old
    uid = new.read_text().strip()

if not old.exits():
    old.write_text(f'{{"user_id": "{uid}"}}')
```

they're nearly (but not quite) the same.

```suggestion
	if os.IsNotExist(errorNew) {
		if os.IsNotExist(errorOld) {
			os.MkdirAll(filepath.Dir(old), 0644)
			data := map[string]interface{}{
				"user_id": id,
			}
			file, _ := json.MarshalIndent(data, "", " ")
			ioutil.WriteFile(old, file, 0644)
		} else {
			jsonFile, jsonErr := os.Open(old)

			if jsonErr == nil {
				byteValue, _ := ioutil.ReadAll(jsonFile)
				var data map[string]interface{}
				json.Unmarshal([]byte(byteValue), &data)
				id = data["user_id"].(string)

				defer jsonFile.Close()
			}
		}

		os.MkdirAll(filepath.Dir(new), 0644)
		ioutil.WriteFile(new, []byte(id), 0644)
	} else {
		dat, _ := ioutil.ReadFile(new)
		id = string(dat[:])
	}
```

DavidGOrtega: @casperdcl your ideal solution is not what we agreed, we were writing the old format in any case with our generated id.

iterative/utils/analytics.go
DavidGOrtega: With the addition of 
```go
if _, ok := os.LookupEnv("CI"); ok {
		return "unknown"
	}
```

I guess we should refactor isCI as 

```go
func IsCI() bool {
	return len(guessCI()) > 0 
}
```

iterative/utils/analytics.go
DavidGOrtega: @casperdcl

iterative/utils/analytics.go
```diff
@@ -0,0 +1,351 @@
...
+	return id.String()
+}
+
+func UserId() string {
+	if IsCI() {
+		ci := guessCI()
+		var rawId string
+
+		if ci == "github" {
+			rawId = os.Getenv("GITHUB_ACTOR")
```
casperdcl, line 168: should be `name login id`, e.g. `Casper da Costa-Luis casperdcl 10780059`

iterative/utils/analytics.go
```diff
@@ -0,0 +1,351 @@
...
+		} else if ci == "gitlab" {
+			rawId = fmt.Sprintf("%s %s %s",
+				os.Getenv("GITLAB_USER_NAME"),
+				os.Getenv("GITLAB_USER_LOGIN"),
+				os.Getenv("GITLAB_USER_ID"))
+		} else if ci == "bitbucket" {
+			rawId = os.Getenv("BITBUCKET_STEP_TRIGGERER_UUID")
```
casperdcl, line 175: ah, nice. Didn't know this was a thing.

casperdcl (changes requested):

iterative/utils/analytics.go
casperdcl: hmm ok - updated suggestion. How about now?

DavidGOrtega: Include the `do-not-track` exception. 👍

iterative/utils/analytics.go
DavidGOrtega: Could you please elaborate a bit more? Its not `GITHUB_ACTOR` the right one?

iterative/utils/analytics.go
DavidGOrtega: @casperdcl that check is not needed. We can not specify the id within the provider. The only way of having `do-not-track` as an id should be reading the new or old credentials generated by other iterative product. If thats the case the file exists and are not going to be written

DavidGOrtega: @casperdcl that check is not needed. We can not specify the id within the provider. The only way of having `do-not-track` as an id should be reading the new or old credentials generated by other iterative product. If thats the case the file exists and are not going to be written or the old credentials are going to be written with `do-not-track` which should be fine

iterative/utils/analytics.go
dacbd: `GITHUB_ACTOR` is the username like: `dacbd`, `casperdcl` , or `DavidGOrtega`

iterative/utils/analytics.go
```diff
@@ -0,0 +1,351 @@
...
+	tpiLogs := d.Get("logs").([]interface{})
+	logs := ""
+	for _, log := range tpiLogs {
+		logs += log.(string)
+	}
+	spot := d.Get("spot").(float64)
+	return map[string]interface{}{
+		"cloud":           d.Get("cloud").(string),
+		"cloud_region":    d.Get("region").(string),
+		"cloud_machine":   d.Get("machine").(string),
```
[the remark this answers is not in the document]
0x2b3bfa0: They can't: this property always refers to either an _instance type_ as documented on cloud providers, or a _generic instance type_ as defined in our code.

iterative/utils/analytics.go
casperdcl: exactly, we don't want just GITHUB_ACTOR, we want a bit more info too... marginally improves security (harder to reverse the hash)

casperdcl: https://github.com/octokit/go-octokit might help :)

casperdcl (approved):
:partying_face: lgtm!

PR #495, opened 2022-04-12
Bump `node` to v16

0x2b3bfa0:
Includes running `go-bindata` 🙃 
```console
cd iterative
go install github.com/go-bindata/go-bindata/...@latest
go-bindata -ignore=README.md -pkg iterative ../environment
```

PR #497, opened 2022-04-12
Replace `go-bindata` with `go:embed`

0x2b3bfa0:
[Available](https://pkg.go.dev/embed) since Go 1.16; much more transparent and easier to maintain than `go-bindata`; motivated by #495


DavidGOrtega (approved):
👍

PR #506, opened 2022-04-14
`runner` Support GCP OIDC credentials 

dacbd:
I am satisfied with the additional testing I have done.


iterative/utils/helpers.go
```diff
@@ -89,16 +91,36 @@ func SetId(d *schema.ResourceData) {
...
+	// from saString, yank this
+	// -----------------------v.........v----------------------
+	// (...serviceAccounts/cml@cml-pulse.iam.gserviceaccount.com:...)
+	atIndex := strings.Index(saString, "@") + 1
+	iamIndex := strings.Index(saString, ".iam.")
+	if atIndex == -1 || iamIndex == -1 {
+		return "", errors.New("[GCP OIDC] Failed to get Project ID from service_account_impersonation_url")
+	}
+	projectID := saString[atIndex:iamIndex]
+	return projectID, nil
```
0x2b3bfa0, line 123: What about having [100 problems](https://xkcd.com/1171/)? 😄 
```suggestion
func GCPCoerceOIDCCredentials(credentialsJSON []byte) (string, error) {
	credentials := make(map[string]interface{})

	if err := json.Unmarshal(credentialsJSON, &credentials); err != nil {
		return "", err
	}

	if url, ok := credentials["service_account_impersonation_url"].(string); ok {
		re := regexp.MustCompile("^https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/.+?@(?P<project>.+).iam.gserviceaccount.com:generateAccessToken$")
		if match := re.FindStringSubmatch(url); match != nil {
			return match[1], nil
		}
		return "", errors.New("failed to get project identifier from service_account_impersonation_url")
	}

	return "", errors.New("unable to load service_account_impersonation_url")
```

iterative/gcp/provider.go
```diff
@@ -324,7 +324,27 @@ func getProjectService() (string, *gcp_compute.Service, error) {
...
+			"CLOUDSDK_CORE_PROJECT",
+			"CLOUDSDK_PROJECT",
+			"GCLOUD_PROJECT",
+			"GCP_PROJECT",
+			"GOOGLE_CLOUD_PROJECT",
+			"GOOGLE_PROJECT",
+		})
+		if coercedProjectID == "" {
+			// last effort to load
+			fromCredentialsID, err := utils.GCPCoerceOIDCCredentials(credentials.JSON)
```
0x2b3bfa0, line 341: It's already too long, but should it be a bit more explicit?
```suggestion
			fromCredentialsID, err := utils.getGCPProjectIDfromOIDCCredentials(credentials.JSON)
```

dacbd: I like explicit, would need that Golang "start func with a capitol letter for 'public'"

0x2b3bfa0: Oops! Edited, I forgot it was being used from another package. 🙈 Should this function live perhaps under `terative/gcp/provider.go` instead?

iterative/utils/helpers.go
dacbd: I do tend to avoid it. What you have here is still pretty readable and tests well. 
/tomorrow I will apply this and run my sanity checks to be safe but I like it.

dacbd:
In theory it could be used in the future by task/other resources? I'll
leave that up to you.

0x2b3bfa0:
I's prefer to keep the scope as local as possible for now, given that #461 may deduplicate[^1] some legacy code, and it may be easier to migrate the relevant portions than to import them from the “utils” [42](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy#The_Answer_to_the_Ultimate_Question_of_Life,_the_Universe,_and_Everything_is_42)–ish package.

[^1]: Blatantly obvious euphemism for obliterate

dacbd:
Works for me, I will de-utils some of these gcp functions.

iterative/gcp/provider.go
```diff
@@ -514,3 +552,46 @@ func getInstanceType(instanceType string, instanceGPU string) (map[string]map[st
...
+		"useraccounts-ro":       "https://www.googleapis.com/auth/cloud.useraccounts.readonly",
+		"useraccounts-rw":       "https://www.googleapis.com/auth/cloud.useraccounts",
+		"userinfo-email":        "https://www.googleapis.com/auth/userinfo.email",
+	}
+	if matchedURL, ok := scopeMap[scope]; ok {
+		return matchedURL
+	}
+	return scope
+}
+func getCanonicalizeServiceScopes(scopes []string) []string {
```
0x2b3bfa0, line 591:
```suggestion
func getCanonicalizedServiceScopes(scopes []string) []string {
```

❓

0x2b3bfa0 (approved):
Looks good to me

PR #516, opened 2022-04-16
Add timestamps to `k8s` logs

0x2b3bfa0:
For consistency with the other backends


dacbd (approved):
lgtm:
https://pkg.go.dev/k8s.io/api/core/v1#PodLogOptions
```
        // If true, add an RFC3339 or RFC3339Nano timestamp at the beginning of every line
	// of log output. Defaults to false.
	// +optional
	Timestamps [bool](https://pkg.go.dev/builtin#bool) `json:"timestamps,omitempty" protobuf:"varint,6,opt,name=timestamps"`
```

PR #517, opened 2022-04-17
Remove `journalctl` header and footer

0x2b3bfa0:
Redundant and not consistent across backends (see Kubernetes)
```
-- Logs begin at Sat 2022-04-16 16:34:38 UTC, end at Sat 2022-04-16 16:36:01 UTC
```


dacbd (approved):
lgtm, tested against a different service I have running and the output looks pretty sane to me.

PR #537, opened 2022-04-26
Set version in `.goreleaser.yml`

0x2b3bfa0:
Cascades to #493

### DIFF
--- a/iterative/utils/analytics.go
+++ b/iterative/utils/analytics.go
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	Version string = "v0.0.0"
+	Version string = "0.0.0"
 )
 
 func getenv(key, defaultValue string) string {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
-          args: release --rm-dist -ldflags="-X 'terraform-provider-iterative/iterative/utils.Version=${GITHUB_REF##refs/tags/v}'"
+          args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   synchronize:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,7 +5,11 @@ builds:
   flags:
     - -trimpath
   ldflags:
-    - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
+    - -s -w
+    - -X terraform-provider-iterative/iterative/utils.Version={{.Version}}
+    - -X main.version={{.Version}}
+    - -X main.commit={{.Commit}}
+    - -X main.date={{.Date}}
   goos:
     # - freebsd
     - windows


iterative/utils/analytics.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	Version string = "v0.0.0"
+	Version string = "0.0.0"
```
DavidGOrtega, line 23: We need to change also the test 

```
func TestVersion(t *testing.T) {
	assert.Equal(t, strings.HasPrefix(Version, "v"), true)
}
```

DavidGOrtega (changes requested):

iterative/utils/analytics.go
0x2b3bfa0: https://github.com/iterative/terraform-provider-iterative/pull/537/commits/f0adf73e218436e734a0b71a4091e8cee76412aa

iterative/utils/analytics_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func TestVersion(t *testing.T) {
-	assert.Equal(t, strings.HasPrefix(Version, "v"), true)
+	assert.Equal(t, "0.0.0", true)
```
DavidGOrtega, line 11: I would do something more flexible, indeed you can test that the ldflags works with 

```sh
go test -ldflags="-X 'terraform-provider-iterative/iterative/utils.Version=v3.0.0'" ./...
```

DavidGOrtega (changes requested):

iterative/utils/analytics_test.go
0x2b3bfa0: A bit of defensive testing, perhaps? If we were using `gereleaser` for testing (hypothetically), it would make sense to check that the configuration is correct. Passing this flag in another place is equally prone to errors, and the only thing we're testing is that Go understands that flag. 🤯

0x2b3bfa0: Voting to “resolve unresolved”

DavidGOrtega: > the only thing we're testing is that Go understands that flag

Not really, It helped me to figure out the right package structure to alter the Version. 
Not for automatised tests but for human checks... I agree that we might not needed it anymore but damn I was terribly helpful

0x2b3bfa0: It would be helpful to check that `goreleaser` honors that option, but testing that is really, _really_ clunky. It may sound a bit lazy, but I would just omit this test. 🐊 

@iterative/cml, opinions?

DavidGOrtega: It fine, lets do not waste time on this.
I would have changed it ot check if version has three dots or something like that
But I guess its fine

0x2b3bfa0: > _if version has three dots_

(Also known as, it's _exactly_ `0.0.0`)

0x2b3bfa0:
Merging as per https://github.com/iterative/terraform-provider-iterative/pull/537#discussion_r858425239; feel free to amend it in the target branch as you see fit.

PR #538, opened 2022-04-26
Add panic handling and background analytics

0x2b3bfa0:
Cascade on #493

* Sends `panic` contents and stack trace as part of analytics
* Runs analytics in the background and waits `Timeout` seconds for them to complete on exit
* Honors the analytics kill switches from the specification
* Renames some environment variables

~[Question: is `?ip_policy=strict` needed?](https://iterativeai.slack.com/archives/C0249LW0HAQ/p1650371913116649?thread_ts=1649179052.132839&cid=C0249LW0HAQ)~ Yes, see https://github.com/iterative/terraform-provider-iterative/pull/538/commits/dc3f90c1904d07851f775fcc74687b4e52690321

~We may want to use a clean context for analytics so it doesn't get cancellation signals from Terraform.~ Yes, see https://github.com/iterative/terraform-provider-iterative/pull/538/commits/f4d153d8cb635f1d44bdb64380f76e1aae6dad11

### DIFF
--- a/iterative/utils/analytics.go
+++ b/iterative/utils/analytics.go
@@ -2,7 +2,6 @@ package utils
 
 import (
 	"bytes"
-	"context"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -241,12 +240,11 @@ func SendJitsuEvent(action string, e error, extra map[string]interface{}) {
 		}
 	}
 
+	go send(JitsuEventPayload(action, e, extra))
 	wg.Add(1)
-	ctx, _ := context.WithTimeout(context.Background(), Timeout)
-	go send(ctx, JitsuEventPayload(action, e, extra))
 }
 
-func send(ctx context.Context, event interface{}) error {
+func send(event interface{}) error {
 	defer wg.Done()
 
 	body, err := json.Marshal(event)
@@ -254,7 +252,7 @@ func send(ctx context.Context, event interface{}) error {
 		return err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, "POST", getenv("TPI_ANALYTICS_ENDPOINT", Endpoint), bytes.NewBuffer(body))
+	req, err := http.NewRequest(http.MethodPost, getenv("TPI_ANALYTICS_ENDPOINT", Endpoint), bytes.NewBuffer(body))
 	if err != nil {
 		return err
 	}
@@ -262,7 +260,8 @@ func send(ctx context.Context, event interface{}) error {
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("X-Auth-Token", getenv("TPI_ANALYTICS_TOKEN", Token))
 
-	resp, err := http.DefaultClient.Do(req)
+	client := &http.Client{Timeout: Timeout}
+	resp, err := client.Do(req)
 	if err != nil {
 		return err
 	}
--- a/iterative/utils/analytics.go
+++ b/iterative/utils/analytics.go
@@ -212,7 +212,7 @@ func JitsuEventPayload(action string, e error, extra map[string]interface{}) map
 
 	err := ""
 	if e != nil {
-		err = reflect.TypeOf(e).Name()
+		err = reflect.TypeOf(e).String()
 	}
 
 	payload := map[string]interface{}{
--- a/iterative/utils/analytics.go
+++ b/iterative/utils/analytics.go
@@ -234,9 +234,9 @@ func JitsuEventPayload(action string, e error, extra map[string]interface{}) map
 }
 
 func SendJitsuEvent(action string, e error, extra map[string]interface{}) {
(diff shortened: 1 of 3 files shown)


iterative/utils/analytics.go
```diff
@@ -22,8 +26,15 @@ import (
 	"golang.org/x/crypto/scrypt"
 )
 
+const (
+	Timeout  = 5 * time.Second
+	Endpoint = "https://telemetry.cml.dev/api/v1/s2s/event?ip_policy=strict"
+	Token    = "s2s.jtyjusrpsww4k9b76rrjri.bl62fbzrb7nd9n6vn5bpqt"
+)
+
```
DavidGOrtega, line 34: I do not get the `ip_policy=strict` issue/craziness. It was already set?!

iterative/utils/analytics.go
```diff
@@ -223,26 +233,59 @@ func JitsuEventPayload(action string, e error, d *schema.ResourceData) map[strin
...
+		if _, ok := os.LookupEnv(prefix + "_NO_ANALYTICS"); ok {
+			logrus.Debugf("analytics: %s_NO_ANALYTICS environment variable is set", prefix)
+			return
+		}
 	}
 
-	if _, ok := os.LookupEnv("DO_NOT_TRACK"); ok {
-		logrus.Debug("Analytics: DO_NOT_TRACK enabled")
-		return
-	}
```
DavidGOrtega, line 234: I prefer the well known `DO_NOT_TRACK` environment variable. If any change should be done our products should use it.

iterative/utils/analytics.go
0x2b3bfa0: It was because of [this conversation](https://iterativeai.slack.com/archives/C0249LW0HAQ/p1650371913116649?thread_ts=1649179052.132839&cid=C0249LW0HAQ).

0x2b3bfa0: Nevermind

iterative/utils/analytics.go
0x2b3bfa0: DVC uses `DVC_NO_ANALYTICS`, and in the specification we have `ITERATIVE_NO_ANALYTICS` as a general alternative. Disclaimer: it was me who wrote it. 😉

0x2b3bfa0: No strong opinions on naming, but please discuss this at the specification level so all the products use the same name.

iterative/utils/analytics.go
```diff
@@ -223,26 +233,59 @@ func JitsuEventPayload(action string, e error, d *schema.ResourceData) map[strin
...
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Auth-Token", getenv("TPI_ANALYTICS_TOKEN", Token))
+
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		logrus.Info("Analytics: failed sending event")
+		return err
+	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
0x2b3bfa0: Yes, and it behaves as expected. Note that `resp.Body.Close()` will only run if `err != nil` and not always.

iterative/utils/analytics.go
DavidGOrtega: yes `DO_NOT_TRACK` is [well known ](https://www.google.com/search?q=site%3Agithub.com+DO_NOT_TRACK)

0x2b3bfa0: https://consoledonottrack.com

iterative/utils/analytics.go
```diff
@@ -223,26 +233,59 @@ func JitsuEventPayload(action string, e error, d *schema.ResourceData) map[strin
...
+			return
+		}
 	}
 
-	if _, ok := os.LookupEnv("DO_NOT_TRACK"); ok {
-		logrus.Debug("Analytics: DO_NOT_TRACK enabled")
-		return
-	}
+	wg.Add(1)
+	ctx, _ := context.WithTimeout(context.Background(), Timeout)
```
dacbd, line 245: Without calling the returned cancel func doesn't this technically leak the context resources if the goroutine finishes before the timeout?

I don't think it much of a real issue since it will generally be terminating right after, ie this is never a long running process.

0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/538/commits/b305ad58e43d8731d1eb04ad41a835ac547f2d97 and getting rid of contexts. You're right about “leaking” the context resources for exactly 5 seconds, even if the goroutine finished a bit before.

0x2b3bfa0: Definitely not an issue, but we weren't using the cancel function anyway, so it's better to get rid of it in favor of the native timeout.

dacbd (approved):
LGTM

PR #546, opened 2022-04-28
_slightly_ "Improve" human readability of start-up script logs

dacbd:


0x2b3bfa0:
Alternatively, [`PS4='tpi:setup.sh: '; set -x`](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) and `fi | sed "s/^/$PS4 /"`

dacbd:
> Alternatively, [`PS4='tpi:setup.sh: '; set -x`](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) and `fi |& sed "s/^/$PS4 /"`

that is arguably cleaner

dacbd:
giving it a try

0x2b3bfa0 (approved):
Looks great to me! I hope you don't mind me preventing you from doing `evil` 😈

0x2b3bfa0:
> _Does one ever master bash-fu?_

I don't think so.

dacbd:
On a side note, I Iike that eval is where you draw the line, but didn't somebody recently say friends don't let friends `curl | sh` I see two cases pretty close by 🤔 😇

PR #549, opened 2022-04-29
Proper GCP - OIDC termination

dacbd:
You can only generate one access token per oidc authorization, so if oidc creds are used we need to pass the access_token to the instance to properly self-terminate.

Closes: https://github.com/iterative/cml/issues/989


casperdcl (approved):
lgtm - but is there any way we can reliably test this? Should we?

dacbd:
> is there any way we can reliably test this? 

Yes, it is something I have been working on

> Should we?

I think so, I have encountered odd edge cases from my previously setup daily test run that I have used to make sure new changes haven't broken any key features or as well as help to identify intermittent issues

PR #554, opened 2022-05-04
Escape dots in GCP OIDC regular expression

0x2b3bfa0:
Thanks, CodeQL! 😅


dacbd:
going test run this branch through my test.

dacbd (approved):
As expected, no issue. Also @0x2b3bfa0 we should do a release as the gcp-oidc fix isn't in one yet.

PR #575, opened 2022-05-12
Adjust some actions triggers

dacbd:
Feels like some easy low hanging fruit, to make docs iterations _slightly_ faster


0x2b3bfa0 (approved):

casperdcl (commented):
I assume codeql doesn't check markdown? Should we add an action that does?

0x2b3bfa0:
We may want to have a separate check for prose.

PR #582, opened 2022-05-17
`runner` tempdir patch

dacbd:
- partially fixes: https://github.com/iterative/cml/issues/1006
 - semi-related: https://github.com/iterative/terraform-provider-iterative/issues/581
 
I think the real fix is probably to run the GHA client in a separate `slice` enforcing resource limits

### DIFF
--- a/iterative/resource_runner.go
+++ b/iterative/resource_runner.go
@@ -344,7 +344,7 @@ export KUBERNETES_CONFIGURATION={{escape .KUBERNETES_CONFIGURATION}}
 {{.runner_startup_script}}
 {{- end}}
 
-HOME="$(mktemp -d)" exec $(which cml-runner || echo $(which cml-internal || echo cml) runner) \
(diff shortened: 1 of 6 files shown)


0x2b3bfa0 (commented):
Silly question: is the default `mktemp` path being mounted as `tmpfs` on the images we use? 🤔

dacbd:
sorry I just saw that the link PR has been merged so I assumed it's been released?

0x2b3bfa0:
As per the relevant milestone, it has not been released yet

0x2b3bfa0:
> _No, as I have seen it survive reboots, however, I believe there is a reaping process that hits the dir. Or there was an existing process already in memory that deleted enough files from disk when it got SIGTERM from rebooting my test._

I guess I don't have enough context. 😅 Looks clear as  to me.

0x2b3bfa0:
If `/tmp` is not mounted as `tmpfs`, what does this pull request fix? 🤔

PR #605, opened 2022-06-10
Add `compute.networks.create` to `gcp` role in permissions doc example `main.tf`

danieljimeneznz:
When I first attempted to deploy this to a newly created Google project this additional permission is required as all `cml` runners are placed in the `iterative` network by default, which will be created by the provider if it doesn't exist already, hence the additional perm.


dacbd (approved):
Good catch, Thanks!

Issue #531, opened 2022-04-24
Local port forwarding


0x2b3bfa0:
It would be nice to have a command for local port forwarding, so users can securely access services running on the `task` from their local machines.

* `aws`, `az` & `gcp` — either [user–mode WireGuard](https://fly.io/blog/ssh-and-user-mode-ip-wireguard) or SSH tunneling
* `k8s` — see [`portforward.go`](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/portforward/portforward.go)

Related to #291

Issue #291, opened 2021-11-24
Common method for command execution


0x2b3bfa0:
It would be nice[^1] to have an `Execute` method to run commands in any of the active instances.

https://github.com/iterative/terraform-provider-iterative/blob/57625b6db38238a18dd10797ac239710c5f94f8a/task/task.go#L37-L47

### Loose ends

* How to deal with parallelism?
   * Run the command on every instance and return a [slice](https://go.dev/ref/spec#Slice_types) of results.
   * Allow users to specify an index in the range `[0, parallelism[` and return an error if there aren't enough instances running to satisfy the requested index.
   * Allow users to list the running instances for a given task and select one of them to run the command on.
   * ~Pick a random instance every time.~ I had to include it. 😂 

### Draft proposal

* `Execute(instance uint16, input string, command ...string) ([]Result, error)`
  * `instance` represents the instance index in the range `[1, parallelism]`; 0 will run the command on all the active instances.
  * `input` optionally, the standard input to feed the remote command with; `""` will disable this function.
  * `command` is the [slice](https://go.dev/ref/spec#Slice_types) of exec–like arguments; “no arguments” can default to `sh` or `bash` if appropriate.
  * `[]Result` is a [slice](https://go.dev/ref/spec#Slice_types) of [`struct`](https://go.dev/ref/spec#Struct_types) containing
     * Standard output
     * Standard error
     * Exit code

* `ExecuteTTY(instance uint16, streams Streams, command ...string) error`
   Same as above but with TTY and interactive streams, so it behaves like an ordinary SSH session. See [`kubectl exec`](https://github.com/kubernetes/kubectl/blob/caeb9274868c57d8a320014290cc7e3d1bcb9e46/pkg/cmd/exec/exec.go#L123-L129) for an example implementation.

[^1]: Not necessary because #237 eliminates the need of SSH by implementing #126

PR #610, opened 2022-06-23
Provide a standalone command-line tool

0x2b3bfa0:


0x2b3bfa0:
@casperdcl, you may want to convert the following commit comments into review comments:
- https://github.com/iterative/terraform-provider-iterative/pull/610#commitcomment-73677283
- https://github.com/iterative/terraform-provider-iterative/pull/610#commitcomment-73677362
- https://github.com/iterative/terraform-provider-iterative/pull/610#commitcomment-73677417

cmd/root.go
```diff
@@ -0,0 +1,160 @@
...
+										viper.Set("workdir", value)
+									}
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
0x2b3bfa0, line 140: For the record, this can be simplified with `viper.Sub("resource.0...")` paths, but I'll have to dig into https://github.com/spf13/viper/issues/1350 first; it doesn't work due to an upstream issue.

cmd/root.go
```diff
@@ -0,0 +1,160 @@
...
+									viper.Set(strings.ReplaceAll(option, "_", "-"), value)
+								}
+							}
+							for _, option := range []string{
+								"tags",
+								"environment",
+							} {
+								if value, ok := options[option]; ok {
+									for _, nestedBlock := range value.([]map[string]interface{}) {
+										viper.Set(option, nestedBlock)
```
dacbd, line 121: I heard you like `\t`

dacbd (changes requested):
Slightly from our DM conversation. It seems detection / running off an existing `main.tf` does not work.

I (git)checked out the branch / ran `make build` / `cp`'d the binary to the project with my existing `main.tf` and tried several variations to run with it.

- `./tpi`
- `./tpi create`
- `./tpi create ./main.tf`
- `./tpi create .`

```cli
dabarnes@ubuntu:~/iterative/tpi-remote-vscode$ ./terraform-provider-iterative create
Error: required flag(s) "cloud" not set
Usage:
  task create [command...] [flags]

Flags:
      --disk-size int                disk size in gigabytes (default -1)
      --environment stringToString   environment variables (default [])
  -h, --help                         help for create
      --image string                 machine image (default "ubuntu")
      --machine string               machine type (default "m")
      --name string                  deterministic name
      --output string                output directory to download
      --parallelism int              parallelism (default 1)
      --permission-set string        permission set
      --script string                script to run
      --spot                         use spot instances
      --tags stringToString          resource tags (default [])
      --timeout int                  timeout (default 86400)
      --workdir string               working directory to upload (default ".")

Global Flags:
      --cloud string    cloud provider
      --log string      log level (default "info")
      --region string   cloud region (default "us-east")

dabarnes@ubuntu:~/iterative/tpi-remote-vscode$ 

```

cmd/root.go
0x2b3bfa0: Do you mean that it's too nested and too ugly? 😄 Blame https://github.com/spf13/viper/issues/1350 for not being able to use the appropriate syntax and flattening everything. 🙃

0x2b3bfa0: I'll have to contribute it upstream one of these days. 😅

0x2b3bfa0:
> _`I create` doesn't print a name and..._
> _without names `I list` is very unhelpful_

Solved with https://github.com/iterative/terraform-provider-iterative/pull/610/commits/88cede1d2ac0b0a9d1f2e663d0e06d2ed0f14a56 and, perhaps, you can try `I create --name=my-experiment-$(uuidgen)` to set human-readable names.

0x2b3bfa0:
> _I can't use `I read` from the same dir_

You should use `I read <name>` where `<name>` can be retrieved with the `I list` command. There is no way for `I` to guess which task you want to `read` because there is no local state.

It's either “read all” or “read chosen” but there is no brain reading involved. 🧠🔌

dacbd:
After meticulously removing different files and the lines from my original `main.tf` I have determined that if you have a `timeout = int` in your `main.tf` it fails

0x2b3bfa0:
Only if it's an expression, as per https://github.com/iterative/terraform-provider-iterative/pull/610#issue-1282657545, right?

0x2b3bfa0:
That's why this pull request updates the documentation to use `86400` everywhere... until I figure out what magic to do upstream for it to work. 😅

0x2b3bfa0:
> _delete states that it read from the config, being "stateless" does read from the config/need anything from it?_

Yes, it needs a few insignificant details like the cloud provider and the region...

cmd/create/create.go
```diff
@@ -0,0 +1,125 @@
+package create
```
dacbd, line 1: for thread/ reduce top comments?

0x2b3bfa0:
> _also there is some fun counting going on there `x/8` or `x/9`_ 😀

[Nobody Expects the Spanish Inquisition!](https://www.youtube.com/watch?v=HqF_nPbX_Ow) 😅 Ideally, those numbers should be calculated algorithmically (https://github.com/iterative/terraform-provider-iterative/issues/290) to avoid off-by-many issues. Anyway, not directly related to this pull request; probably worth a separate discussion.

dacbd (commented):
It would be nice for `tpi create` to send the name to stdout for better scripting.

`Reading configuration from /home/dabarnes/iterative/test/main.tf` +> stderr
and:
`INFO[0016] Successfully created task: tpi-3kvxtq1k267j1-12kbren5-49f0jokh` > just `[name]` +> stdout

cmd/create/create.go
0x2b3bfa0: > _It would be nice for `I create` to send the name to stdout for better scripting._

Our current logging system is not quite friendly to scripts: https://github.com/iterative/terraform-provider-iterative/pull/423#pullrequestreview-901140497 but yes, that's a good idea; I'll hack something together.

0x2b3bfa0: * https://github.com/iterative/terraform-provider-iterative/pull/610/commits/e4a01d4905993b1731a4396db0fc1ee967ed4092 causes `I` to complain only for configuration parsing errors, and be silent otherwise.
* https://github.com/iterative/terraform-provider-iterative/pull/610/commits/aa0c66e1f19f7cb35ee95fae0c955020e41a038a prints the task name to standard output using a lowly `fmt.Println` 😄

cmd/root.go
0x2b3bfa0: Duplicate of https://github.com/iterative/terraform-provider-iterative/pull/610#discussion_r905888407 😄

0x2b3bfa0:
> _The is code for tpi stop but it doesn't do anything / cmd not found_

Argh! Sorry, having the bad habit of using `init()` to automatically register commands, I forgot to add them. Fixed with https://github.com/iterative/terraform-provider-iterative/pull/610/commits/48180755d8e4e3416e20618eabe7089179c225ac.

0x2b3bfa0:
> _There is a theme of stdout vs stderr tpi list, tpi read, tpi delete do "nothing" from this pov (delete is probably fine though?)_

Can you please elaborate?

0x2b3bfa0:
> _Can we present this better? like status: waiting for instance / instance running?_

Sounds like you may have overlooked the very last line of `I read` (?) 🤔 It displays colorful status information like queued/running/failed/succeeded.

Captura de pantalla 2022-06-29 a las 1 17 04

dacbd:
> > There is a theme of stdout vs stderr `I list`, `I read`, `I delete` do "nothing" from this pov (delete is probably fine though?)
>
> Can you please elaborate?

sorry I explained this better in my first version of this review before it was violently lost 😞 ...

from the point of view of these commands being able to nicely work as part of a larger system most output is all sent to stderr
`./tpi read tpi-3u75pr8sir0hd-t6802j39-3qwi6brx 2>/dev/null` 

or perhaps something simpler you can see what I feel is counterintuitive:
```
tpi read [name] > [name].log
wc --bytes [name].log
```

to additionally elaborate, the log parsing is inconsistent for machine parsing, and actually playing with it more I'm not even sure what weirdness this logging library is trying to do...

I cant go `./tpi read tpi-3u75pr8sir0hd-t6802j39-3qwi6brx |& awk 'match($0, /LOG 0 >> /){print $0}' > my_scripts_logs.log` and have this do what I would expect.

I would prefer for equal treatment in the interpretation of `\n`  note the difference between `./tpi read [name]` vs `./tpi read [name] |& cat` and this is a frivolous use of `cat` because of the above note where `./tpi read [name] > file` results in an empty file.

0x2b3bfa0:
I wholeheartedly agree:

1. Command output differs between teletypes and otherwise
    * E.g. running `command` on your terminal is different to running `command |& cat`
    * None of the two formats are good for scripting
2. Command output goes to standard error, making parsing a lot more difficult
3. There is a lot to improve; e.g. be able to output just raw logs, or just the status, ...

Not sure if that's on scope for this pull request, but is definitely worth addressing before making this command-line tool generally available-

dacbd (approved):
Taking a note of our previous comments I think this works much better or at least feels much nicer than the terraform flow.  As a first pass 👍!

What are the next steps?

0x2b3bfa0:
#### Next steps, copied from https://github.com/iterative/terraform-provider-iterative/pull/610#issue-1282657545
* https://github.com/iterative/terraform-provider-iterative/issues/291
* https://github.com/iterative/terraform-provider-iterative/issues/531
#### Next steps (user experience)
* [ ] Improve command-line interface
  * [ ] Provide sensible output formats for TTY and non-TTY
  * [ ] Separate commands to list logs & general status[^1]
  * [ ] Allow users to specify custom identifiers with a random suffix

[^1]: With a special case for $parallelism > 1$ where users can filter logs for a single task.

cmd/create/create.go
```diff
@@ -0,0 +1,134 @@
...
+			},
+		},
+		Parallelism: uint16(1),
+	}
+
+	if o.Spot {
+		cfg.Spot = common.Spot(common.SpotEnabled)
+	} else {
+		cfg.Spot = common.Spot(common.SpotDisabled)
+	}
```
DavidGOrtega, line 102:
```suggestion
        cfg.Spot = common.Spot(common.SpotDisabled)
	if o.Spot {
		cfg.Spot = common.Spot(common.SpotEnabled)
	} 
```

cmd/create/create.go
```diff
@@ -0,0 +1,134 @@
...
+
+	id := common.NewRandomIdentifier()
+
+	if o.Name != "" {
+		if identifier, err := common.ParseIdentifier(o.Name); err == nil {
+			id = identifier
+		} else {
+			id = common.NewIdentifier(o.Name)
+		}
+	}
```
DavidGOrtega, line 112:
```suggestion
	if o.Name != "" {
	        id = common.NewIdentifier(o.Name)
		if identifier, err := common.ParseIdentifier(o.Name); err == nil {
			id = identifier
		}
	}
```

iterative/resource_task.go
```diff
@@ -369,22 +368,26 @@ func resourceTaskBuild(ctx context.Context, d *schema.ResourceData, m interface{
...
-		} else if identifier := os.Getenv("BITBUCKET_STEP_TRIGGERER_UUID"); identifier != "" {
-			name = identifier
+	id, err := common.ParseIdentifier(d.Id())
+	if err != nil {
+		if name := d.Get("name").(string); name != "" {
+			if newId, err := common.ParseIdentifier(name); err == nil {
+				id = newId
+			} else {
+				id = common.NewIdentifier(name)
+			}
```
DavidGOrtega, line 378:
```suggestion
			id = common.NewIdentifier(name)
			if newId, err := common.ParseIdentifier(name); err == nil {
				id = newId
			}
```

Also... I have seen this code before 🤔

DavidGOrtega (approved):
Nitpicks

cmd/create/create.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/610/commits/d3d544e1fc395bb6b965e5ab5297d91e75e8dad7

cmd/create/create.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/610/commits/d3d544e1fc395bb6b965e5ab5297d91e75e8dad7

iterative/resource_task.go
0x2b3bfa0: Yes, sounds familiar; I'll RM [^1] for now, though. 😅 

[^1]: Repeat myself.

DavidGOrtega:
@0x2b3bfa0 happy to merge?

0x2b3bfa0:
@DavidGOrtega :shipit:

PR #614, opened 2022-06-23
`runner` setup script typo

dacbd:
discovered during some spot instance testing 🙈

### DIFF
--- a/environment/setup.sh
+++ b/environment/setup.sh
@@ -1,7 +1,7 @@
 #/bin/sh
 FILE=/var/log/cml_stack.log
 if [ ! -f "$FILE" ]; then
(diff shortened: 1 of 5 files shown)


0x2b3bfa0 (approved):
Argh! 🤦🏼‍♂️

DavidGOrtega:
I never look at it even. That code was added long time ago. I guess that `ecr_helper` was not working at all

dacbd:
indeed I think when I was testing it I had it differently than for the merge, condensed it for readability and it got missed. (thankfully 
 I added so if it failed (as it often would take 10 mins to download for gcp) it wouldn't matter)

DavidGOrtega (approved):

PR #621, opened 2022-07-05
Wait for `apt` to finish

0x2b3bfa0:


environment/setup.sh
```diff
@@ -1,14 +1,15 @@
...
+PS4='tpi:setup.sh: '
+set -x
 
-  sudo apt remove unattended-upgrades
-  systemctl disable apt-daily-upgrade.service
+export DEBIAN_FRONTEND=noninteractive
+echo "APT::Get::Assume-Yes \"true\";" | sudo tee -a /etc/apt/apt.conf.d/90assumeyes
 
+sudo apt remove unattended-upgrades
+systemctl disable apt-daily-upgrade.service
```
dacbd, line 9: do you need sudo to disable services?

0x2b3bfa0: The whole script runs as root

dacbd:
update golden refs for tests?

dacbd (changes requested):
does not have the intended effect. gets stuck in a sleep loop.
my quick test shows that file (`/var/lib/dpkg/lock`) as always present but we probably want something more like `lsof /var/lib/dpkg/lock`

0x2b3bfa0:
Major blunder. 🤦🏼‍♂️ It uses a `fnctl` lock.

iterative/testdata/script_template_cloud_gcp.golden
```diff
@@ -48,6 +49,8 @@ sudo tee /usr/bin/cml.sh << 'EOF'
 export GOOGLE_APPLICATION_CREDENTIALS_DATA=''
 export CML_GCP_ACCESS_TOKEN=''
 
+while ! lsof /var/lib/dpkg/lock; do sleep 1; done
```
dacbd, line 52:
```suggestion
while lsof /var/lib/dpkg/lock; do sleep 1; done
```

dacbd (changes requested):
Also not having the desired effect, caught in a sleep loop.
```
● cml.service
     Loaded: loaded (/etc/systemd/system/cml.service; enabled; vendor preset: enabled)
     Active: active (running) since Tue 2022-07-05 22:17:24 UTC; 7min ago
   Main PID: 14984 (cml.sh)
      Tasks: 2 (limit: 19190)
     Memory: 1.6M
     CGroup: /system.slice/cml.service
             ├─14984 /bin/sh /usr/bin/cml.sh
             └─19491 sleep 1

Jul 05 22:17:24 cml-qoyvzp8y43 systemd[1]: Started cml.service.
```

```
# with no apt process running
sudo lsof /var/lib/dpkg/lock
echo $?
# 1

# with apt process running
sudo lsof /var/lib/dpkg/lock
echo $?
# 0
```

I believe we want to drop the ` ! `

dacbd (approved):
perhaps a patch release?

PR #631, opened 2022-08-01
`task` Fix id-generation

dacbd:
:hear_no_evil: Much sad! 8 hrs of debugging for such a small fix :see_no_evil:

my day's TLDR:

Any id/name greater than 28 aka: `maximumLongLength-shortLength-uint32(len("tpi---"))` resulted in improperly generating the digest and so the inferred bucket name was different from `terraform create` to `terraform refresh` which results in the follow error: 
image

Discovered via some BB testing because below  Bitbucket CI env is a uuid and thus `>28` you can also replicate the error by setting a `name` for your task that is also `>28` chars:
https://github.com/iterative/terraform-provider-iterative/blob/c3f1c08c60d32b954e8e23b23feae2e7431100c5/iterative/resource_task.go#L371-L388

### DIFF
--- a/task/common/identifier.go
+++ b/task/common/identifier.go
@@ -41,7 +41,7 @@ func NewRandomIdentifier() Identifier {
 
 func (i Identifier) Long() string {
 	name := normalize(string(i), maximumLongLength-shortLength-uint32(len("tpi---")))
-	digest := hash(string(i), shortLength/2)
+	digest := hash(name, shortLength/2)
 
 	return fmt.Sprintf("tpi-%s-%s-%s", name, digest, hash(name+digest, shortLength/2))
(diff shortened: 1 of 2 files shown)


task/common/identifier.go
```diff
@@ -41,7 +41,7 @@ func NewRandomIdentifier() Identifier {
 
 func (i Identifier) Long() string {
 	name := normalize(string(i), maximumLongLength-shortLength-uint32(len("tpi---")))
-	digest := hash(string(i), shortLength/2)
```
tasdomas, line 44: Would it make sense to add a test that reproduces this issue?

tasdomas (approved):

Issue #559, opened 2022-05-06
`task` implement `permission_set` for azure


dacbd:
For additional context see:
* #550 
* #501
* https://github.com/iterative/terraform-provider-iterative/issues/235

0x2b3bfa0:
#### To be documented

* https://github.com/iterative/terraform-provider-iterative/pull/632#pullrequestreview-1076595190

Issue #235, opened 2021-11-02
```instance_permission_set``` support for azure


DavidGOrtega:
Add support for permission set in Azure.

PR #632, opened 2022-08-02
Add support for instance-permission-set in azure

tasdomas:
This PR addresses #559 

The instance-permission-set is parsed as a comma-separated list of ARM resource ids in the form
'/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{identityName}'.

### DIFF
--- a/iterative/azure/provider.go
+++ b/iterative/azure/provider.go
@@ -16,6 +16,8 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	azresources "terraform-provider-iterative/task/az/resources"
 )
 
 //ResourceMachineCreate creates AWS instance
@@ -29,11 +31,15 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	customData := d.Get("startup_script").(string)
 	region := GetRegion(d.Get("region").(string))
 	instanceType := getInstanceType(d.Get("instance_type").(string), d.Get("instance_gpu").(string))
-	userAssignedIdentities := getUserAssignedIdentityMap(d.Get("instance_permission_set").(string))
 	keyPublic := d.Get("ssh_public").(string)
 	hddSize := int32(d.Get("instance_hdd_size").(int))
 	spot := d.Get("spot").(bool)
 	spotPrice := d.Get("spot_price").(float64)
+
+	userAssignedIdentities, err := getUserAssignedIdentityMap(d.Get("instance_permission_set").(string))
+	if err != nil {
+		return err
+	}
 
 	metadata := map[string]*string{}
 	for key, value := range d.Get("metadata").(map[string]interface{}) {
@@ -367,9 +373,9 @@ func getVMClient(subscriptionID string) (compute.VirtualMachinesClient, error) {
 
 // getUserAssignedIdentityMap generates a map of user-assigned identity values from a comma-separated
 // list of ARM resource ids.
-func getUserAssignedIdentityMap(identitiesRaw string) map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue {
+func getUserAssignedIdentityMap(identitiesRaw string) (map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue, error) {
 	if len(strings.TrimSpace(identitiesRaw)) == 0 {
-		return nil
+		return nil, nil
 	}
 	identities := strings.Split(identitiesRaw, ",")
 	identityMap := map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
@@ -378,12 +384,15 @@ func getUserAssignedIdentityMap(identitiesRaw string) map[string]*compute.Virtua
 		if len(identity) == 0 {
 			continue
 		}
+		if err := azresources.ValidateARMID(identity); err != nil {
+			return nil, err
+		}
 		identityMap[identity] = &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
 	}
 	if len(identityMap) == 0 {
-		return nil
+		return nil, nil
 	}
-	return identityMap
+	return identityMap, nil
 }
 
 //GetRegion maps region to real cloud regions
--- a/iterative/azure/provider.go
+++ b/iterative/azure/provider.go
@@ -203,6 +203,7 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 		Tags: metadata,
 		Identity: &compute.VirtualMachineIdentity{
 			UserAssignedIdentities: userAssignedIdentities,
+			Type:                   compute.ResourceIdentityTypeSystemAssignedUserAssigned,
 		},
 		Location: to.StringPtr(region),
 		VirtualMachineProperties: &compute.VirtualMachineProperties{
--- a/iterative/azure/provider.go
+++ b/iterative/azure/provider.go
@@ -29,6 +29,7 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	customData := d.Get("startup_script").(string)
 	region := GetRegion(d.Get("region").(string))
 	instanceType := getInstanceType(d.Get("instance_type").(string), d.Get("instance_gpu").(string))
+	userAssignedIdentities := getUserAssignedIdentityMap(d.Get("instance_permission_set").(string))
 	keyPublic := d.Get("ssh_public").(string)
 	hddSize := int32(d.Get("instance_hdd_size").(int))
 	spot := d.Get("spot").(bool)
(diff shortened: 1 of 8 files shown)


dacbd:
You got https://github.com/iterative/terraform-provider-iterative/issues/235 ✔️ 😉 
#559 would live here: https://github.com/iterative/terraform-provider-iterative/tree/master/task/az

tasdomas:
@dacbd ah, I see - I guess I misunderstood the relation between `tpi` and `task`

dacbd:
> @dacbd ah, I see - I guess I misunderstood the relation between `tpi` and `task`

no worries they both needed to be done at some point 👍

task/az/resources/data_source_permission_set.go
```diff
@@ -3,12 +3,17 @@ package resources
...
+	"regexp"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-30/compute"
 
 	"terraform-provider-iterative/task/az/client"
 )
 
+// validateARMID is a regular expression for validating user-assigned identity ids.
+var validateARMID = regexp.MustCompile("/subscriptions/.*/resourceGroups/.*/providers/Microsoft.ManagedIdentity/userAssignedIdentities/.*")
```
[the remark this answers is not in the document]
tasdomas: I think it makes sense to do that. But first I need to figure out why the smoke tests are failing

task/az/task.go
```diff
@@ -110,48 +110,59 @@ type Task struct {
...
-	logrus.Info("[8/10] Creating VirtualMachineScaleSet...")
-	if err := t.Resources.VirtualMachineScaleSet.Create(ctx); err != nil {
-		return err
-	}
-	logrus.Info("[9/10] Uploading Directory...")
-	if t.Attributes.Environment.Directory != "" {
-		if err := t.Push(ctx, t.Attributes.Environment.Directory); err != nil {
+	// Define a list of steps to execute. If we adopt this pattern, we
+	// can declare a named type for the step definition.
+	steps := []struct {
```
DavidGOrtega, line 115: Is this change needed in this PR? I find it a bit out of the scope

iterative/azure/provider.go
```diff
@@ -360,6 +364,27 @@ func getVMClient(subscriptionID string) (compute.VirtualMachinesClient, error) {
 	return client, err
 }
 
+// getUserAssignedIdentityMap generates a map of user-assigned identity values from a comma-separated
+// list of ARM resource ids.
+func getUserAssignedIdentityMap(identitiesRaw string) map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue {
```
DavidGOrtega, line 369: This code is very similar to 

```
identities := strings.Split(ps.Identifier, ",")
	identityMap := map[string]*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue{}
	for _, identity := range identities {
		identity = strings.TrimSpace(identity)
		if identity == "" {
			continue
		}
		if !validateARMID.MatchString(identity) {
			return fmt.Errorf("invalid user-assigned identity id: %q", identity)
		}
		identityMap[identity] = &compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue{}
	}
	if len(identityMap) == 0 {
		ps.Resource = nil
		return nil
	}
	ps.Resource = &compute.VirtualMachineScaleSetIdentity{
		UserAssignedIdentities: identityMap,
```

are they different in scope? Could them be using a common function?

tasdomas: They are pretty much the same thing, but I've not yet found a good place to put common azure-specific functionality. Also, there's the subtle difference that in one place we're creating a map of `*compute.VirtualMachineIdentityUserAssignedIdentitiesValue` and in the other - a map of `*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue`. 

DRY is a nice principle in theory, in practice it comes with its own cost.

task/az/task.go
DavidGOrtega: Agree, but why not follow up in technical debt issue? Just only to not loose the focus when reviewing it.

iterative/azure/provider.go
DavidGOrtega: Im not that picky but fi there is an easy way to just call a generic map creator would be awesome.

task/az/task.go
dacbd: it looks like https://github.com/iterative/terraform-provider-iterative/pull/632/commits/aaa06e9db2dd964870bc39749e8a8a609deb7c96 can pretty easily cherry-picked to another PR and rebased out?

tasdomas: Good point, will do just that - thank you!

0x2b3bfa0 (commented):
Is providing several user-assigned identities a common use case? 🤔

dacbd:
I believe this is also funcationally unique to az, gcp/aws, barring common terminology, allow multiple policies to be attached to a single entity and one entity to be assigned to the instance.

0x2b3bfa0 (commented):
Merge #645 first?

0x2b3bfa0 (commented):
# Test

I'm getting the following error when performing the test described below, although it's probably just me doing something wrong or the `az` command-line tool behaving unexpectedly. Can you please take a look?

```
ERROR: Failed to connect to MSI. Please make sure MSI is configured correctly. 
Get Token request returned http error: 400, reason: Bad Request 
```

## Create a user-assigned identity

Write the code below to a file named `main.tf` and run `terraform init && terraform apply` to create the necessary resources:

```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~> 2.98.0"
    }
  }
}

provider "azurerm" {
  features {}
}

data "azurerm_subscription" "current" {}

resource "azurerm_resource_group" "example" {
  name     = "example"
  location = "westus"
}

resource "azurerm_user_assigned_identity" "example" {
  resource_group_name = azurerm_resource_group.example.name
  location            = azurerm_resource_group.example.location
  name                = "example"
}

resource "azurerm_role_assignment" "example" {
  principal_id       = azurerm_user_assigned_identity.example.principal_id
  role_definition_id = azurerm_role_definition.task.role_definition_resource_id
  scope              = data.azurerm_subscription.current.id
}

resource "azurerm_role_definition" "example" {
  name  = azurerm_user_assigned_identity.example.name
  scope = data.azurerm_subscription.current.id
  permissions {
    actions = ["Microsoft.Compute/virtualMachineScaleSets/read"]
  }
}

output "permission_set" {
  value = azurerm_user_assigned_identity.example.id
}
```

> **Note**
> Permissions in this role definition serve just as an example; users should customize them as they wish

## Create a test task using that identity

```bash
./terraform-provider-iterative create\
 --script='curl --location https://aka.ms/InstallAzureCLIDeb | bash && az login --identity && az vmss list'\
 --permission-set=$(terraform output --raw permission_set)\
 --region=westus\
 --cloud=az
```

> **Note**
> If running manually, you may want to replace `$(terraform output --raw permission_set)` with the permission set value, e.g. `/subscriptions/00000000-0000-0000-0000-000000000000
/resourceGroups/example/providers/Microsoft.ManagedIdentity/userAssignedIdentities/example`

---

Sorry for the late review. 🙈

0x2b3bfa0:
🔔  @tasdomas

tasdomas:
> bell @tasdomas
Yeah - looking into this

tasdomas:
The issue is two-fold here:
1. the command line permission set value was not being propagated to the task (fixed)
2. azure's permission system is complex - it's not enough to add a UMI with the necessary assigned roles (still working on this one)

tasdomas:
> # Test
> 
> I'm getting the following error when performing the test described below, although it's probably just me doing something wrong or the `az` command-line tool behaving unexpectedly. Can you please take a look?
> 
> ```
> ERROR: Failed to connect to MSI. Please make sure MSI is configured correctly. 
> Get Token request returned http error: 400, reason: Bad Request 
> ```

So, this took a while to figure out, there were multiple issues:
1. the `permission_set` flag was not propagated to the task definition (fixed)
3. `az login --identity` will login with the machine's system-assigned identity by default (if it exists), the user-assigned identity needs to be explicitly specified to login with: `az login --identity -u {UMI-id}`. This is far from being intuitive, but we have limited ways of working around this **[1]**.


## [1] 

It appears the `az login --identity` logs in with the system assigned identity, if it exists, or the user assigned identity if only one is set. Currently, when allocating a machine, we maintain the system assigned identity for the scale set, so any user assigned identities need to be specifically logged-in with.

We could assign no system-managed identity if at least one UMI is specified. But this would make `permission_set` modify the allocation instead of augmenting it.

0x2b3bfa0 (approved):
Updated https://github.com/iterative/terraform-provider-iterative/pull/632#pullrequestreview-1076595190 with working instructions; now it works as expected. 🎉

0x2b3bfa0:
With regard to the need of specifying an user I would suggest to simplify user's lives a bit more:
1. Don't accept a comma-separated list of identities; users will have to end up picking a single one to log in.
2. Expose the `permission_set` through an environment variable so users can run `az login --identity --username=$IDENTITY` or similar.

0x2b3bfa0:
> _we maintain the system assigned identity for the scale set_

What is it useful for? Maybe we can just go ahead and leave only the user-managed identity? E.g.

> _We could assign no system-managed identity if at least one UMI is specified. But this would make permission_set modify the allocation instead of augmenting it._

🤔

0x2b3bfa0:
Merging with (irrelevant) `aws` tests failing. Resources, resources... 😅

PR #639, opened 2022-08-05
Parse nested maps from viper configuration

### LINKED ISSUE #638: Fix broken parsing of HCL with maps
The standalone command-line tool is unable to properly parse map fields like `environment` because the code assumed that all the options can be represented as a single string.

0x2b3bfa0:
Closes #638; convoluted workaround to keep `cobra` instead of `viper` as the primary schema.

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -64,8 +64,8 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
 	for name, value := range o.Environment {
 		name = strings.ToUpper(name)
 		variables[name] = nil
-		if value != "" {
-			variables[name] = &value
+		if copy := value; value != "" {
+			variables[name] = &copy
 		}
 	}
 
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -62,6 +62,7 @@ func New(cloud *common.Cloud) *cobra.Command {
 func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) error {
 	variables := make(map[string]*string)
 	for name, value := range o.Environment {
+		name = strings.ToUpper(name)
 		variables[name] = nil
 		if value != "" {
 			variables[name] = &value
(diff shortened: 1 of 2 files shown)


dacbd:
Can you help my brain and give an example of what was broken and now fixed by this? I don't quite follow the info in the linked issue? 

this is broken?
```hcl
environment = { THINIG1 = "", THING2 = "" }
```
but this isn't?
```hcl
environment = { THINIG1 = "" }
```

0x2b3bfa0:
Both were broken. The standalone command-line tool ignored all the map-like attributes from a `main.tf` file.

dacbd (changes requested):
It is improperly assigning values.

with:
```hcl
terraform {
  required_providers {
    iterative = { source = "github.com/iterative/iterative" }
  }
}
provider "iterative" {}
resource "iterative_task" "test" {
  cloud = "aws"
  machine = "t3.small"
  disk_size =-1
  storage {
    workdir = ""
    output = ""
  }
  environment = { 
    GITHUB_PAT= "",
    THING1 = "one", 
    THING2 = "two", 
    THING_ONE = "three",
    THING_TWO = "four"
  }
  script = <<-END
    #!/bin/bash
    env
  END
}
```

running the cli results with:
```
THING_TWO=two
GITHUB_PAT=*** #correctly set
THING_ONE=two
THING1=two
THING2=two
```

cmd/create/create.go
```diff
@@ -62,6 +62,7 @@ func New(cloud *common.Cloud) *cobra.Command {
 func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) error {
 	variables := make(map[string]*string)
 	for name, value := range o.Environment {
+		name = strings.ToUpper(name)
 		variables[name] = nil
 		if value != "" {
 			variables[name] = &value
```
0x2b3bfa0, line 68:
```suggestion
		if copy := value; value != "" {
			variables[name] = &copy
```

The issue you describe wasn't introduce by this pull request, but... 🤦🏼‍♂️

dacbd (approved):

PR #642, opened 2022-08-08
Slight refactoring of machine script generation

tasdomas:
Use embedded template instead of format string.
Add test.
Remove use of map pointer - golang maps are already pointers to a maps structure and can be nil.

### DIFF
--- a/task/aws/task.go
+++ b/task/aws/task.go
@@ -221,7 +221,9 @@ func (t *Task) Delete(ctx context.Context) error {
 			}
 		}
 		logrus.Info("[2/8] Emptying Bucket...")
-		if err := machine.Delete(ctx, (*t.DataSources.Credentials.Resource)["RCLONE_REMOTE"]); err != nil && err != common.NotFoundError {
+
+		// TODO: t.DataSources.Credentials.Resource may be nil, check first.
+		if err := machine.Delete(ctx, t.DataSources.Credentials.Resource["RCLONE_REMOTE"]); err != nil && err != common.NotFoundError {
 			return err
 		}
 	}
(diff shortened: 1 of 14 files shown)


0x2b3bfa0:
Have you considered using [`github.com/sebdah/goldie`](https://pkg.go.dev/github.com/sebdah/goldie) for template tests, like in other places of this project? E.g. 

https://github.com/iterative/terraform-provider-iterative/blob/142cd5b8b1aba9b0deb746b9fc5cfe3b6fecdd39/iterative/testdata/script_template_cloud_aws.golden#L1

task/aws/task.go
```diff
@@ -221,7 +221,9 @@ func (t *Task) Delete(ctx context.Context) error {
 			}
 		}
 		logrus.Info("[2/8] Emptying Bucket...")
-		if err := machine.Delete(ctx, (*t.DataSources.Credentials.Resource)["RCLONE_REMOTE"]); err != nil && err != common.NotFoundError {
+
+		// TODO: t.DataSources.Credentials.Resource may be nil, check first.
```
0x2b3bfa0, line 225: Probably a dumb question, but in which case[s] may `t.DataSources.Credentials.Resource` be `<nil>`? :think:

tasdomas: I bulk added that comment when before I refactored `Credentials.Resource` to not be a pointer to a map, forgot to remove it.

In this case `Resource` can be nil if the data source's `Read` returns an error or is never called. If those were write attempts I'd add `nil` checks here anyway, since intialization of the map is significantly distant from this code for it to be considered external. But go's maps are read-safe even when unallocated, so we're fine.

0x2b3bfa0:
Merging with AWS red because of capacity issues; works as expected.

PR #643, opened 2022-08-10
Use named fields instead of anonymous fields in common.Task

tasdomas:
Legitimate use cases for anonymous fields are few and far between. This is not one of them - we're not accessing properties of the anonymous fields directly.


dacbd (approved):

0x2b3bfa0 (approved):
`:ish:` `:ish:`

PR #651, opened 2022-09-02
Support for existing storage containers in aws and gcp.

tasdomas:
This pull request contains a proposal to use gomock-generated mocks for testing interactions with
cloud APIs in unit tests.


task/common/values.go
```diff
@@ -41,7 +41,15 @@ type Event struct {
 	Code        string
 	Description []string
 }
-type Storage struct{}
+type RemoteStorage struct {
+	// Container stores the id of the container to be used.
+	Container string
+	// Path stores the subdirectory inside the container.
+	Path string
```
0x2b3bfa0, line 48: Can we merge those two so users can specify either `storage` or `storage/prefix` with a single string?

task/common/values.go
```diff
@@ -41,7 +41,15 @@ type Event struct {
...
+	// Config stores provider-specific configuration keys for accessing the pre-allocated
+	// storage container.
+	Config map[string]string
```
0x2b3bfa0, line 51: Maybe too much flexibility? I'd assume that no special options are needed to connect and, if they were, users could just provide a full [`rclone` connection string](https://rclone.org/docs/#connection-strings), removing the maintenance burden on our side.

task/common/resource.go
```diff
@@ -0,0 +1,28 @@
...
+	Delete(ctx context.Context) error
+}
+
+// StorageCredentials is an interface implemented by data sources and resources
+// that provide access to cloud storage buckets.
+type StorageCredentials interface {
+	// ConnectionString returns the connection string necessary to access
+	// an S3 bucket.
+	ConnectionString(ctx context.Context) (string, error)
+}
```
0x2b3bfa0, line 28: Can we assume that there is no need for separate credentials and the storage can be accessed with the same credentials provided to manage the rest of resources? In cases where different credentials are needed (e.g. Microsoft Azure), they can be retrieved using the main credentials:

https://github.com/iterative/terraform-provider-iterative/blob/66e18e4dcacc798b7d9a33a686fb991be4c37eda/task/az/resources/resource_storage_account.go#L70

task/common/resource.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+)
+
+// BlobContainer defines the interface implemented by blob container resources
+// across cloud providers.
+type BlobContainer interface {
+	Resource
+	// Identifier returns the identifier of the container.
+	Identifier() string
+}
```
0x2b3bfa0, line 13: Where are we using this interface? 🤔

task/aws/resources/data_source_bucket_test.go
```diff
@@ -0,0 +1,58 @@
+package resources_test
```
0x2b3bfa0, line 1: Tests are nice, although mocking cloud infrastructure is rather risky. Maybe we can rely on `rclone` to use functionality that it's already tested on their side or can be tested locally with the same API as cloud resources?

task/common/values.go
tasdomas: It depends on what we're aiming at. Rclone connection strings hardly make it easy to configure.

The difference is between
```
  storage {    
    container = "tpi-test"
    container_path = "directory"
    container_opts {
      az_storage_account = "storage-account"
      az_storage_account_key = "storage-account-key"
    }
  }
```
and
```
  storage {
    container = ":azureblob,account='storage-account',key='storage-account-key':tpi-test/directory"
  }
```

I'm not sure the latter is easier, especially if we consider the target users of tpi.

task/common/values.go
tasdomas: I'm hesistant to do so - it feels like separating the two explicitly is clearer. Especially since we'll most likely also need to add a `container-type` field to support pre-allocated buckets in k8s.

task/common/values.go
0x2b3bfa0: More precisely, the difference would be between

```
  storage {    
    container = "tpi-test"
    container_path = "directory"
    container_opts {
      az_storage_account = "storage-account"
      az_storage_account_key = "storage-account-key"
    }
  }
```

and

```
  storage = ":azureblob,account='storage-account',key='storage-account-key':tpi-test/directory"
```

0x2b3bfa0: None of the variants sound particularly easy to me, although the latter is already well docuemnted by `rclone` and, despite being a lengthy string, is fairly readable to my mind.

task/common/resource.go
tasdomas: Ah, good catch - this is a left-over from a previous implementation attempt.

task/common/values.go
0x2b3bfa0: In Kubernetes, this field would probably take a user-specified StorageClass, which is a single string. 🤔 What else could it take?

task/aws/resources/data_source_bucket_test.go
tasdomas: I don't see how mocking cloud apis (not infrastructure) is riskier than having no tests at all.

My current proposal is twofold:
1. Each resource defines limited interfaces (implemented by corresponding cloud apis) that include only the methods actually used.
2. We generate mocks for those interfaces and use them in tests to ensure conditions are handled correctly and api calls are made in the right order.

task/common/values.go
0x2b3bfa0: > _I'm not sure the latter is easier, especially if we consider the target users_

That's a double-edged reason: if our target users are on their own, they probably won't need to specify a custom storage and, if they do, it probably won't require any options beyond the name and path (prefix). 😸 

Note that credentials usually match (or can be retrieved by using) the main credentials used to create resources; see https://github.com/iterative/terraform-provider-iterative/pull/651#discussion_r962436130 for details.

And those users who really need custom options on their connection strings are probably part of a team with opinions, and people having those opinions may as well read `rclone` documentation. 😈 Worst-case scenario, we can provide an interactive `rclone` string builder as part of our documentation.

task/common/resource.go
tasdomas: So, using cloud credentials to retrieve azure storage credentials will only work if the storage is owned by the same azure account. But that is not necessarily the case. I was thinking of also supporting azure sas url's which allow granting access to storage containers by sharing a single url (albeit a lengthy one).

The other side of the issue (and probably the main reason why I moved the connection string generation out to this interface) is separation of concerns. The container resource/data source is the right component to handle generation of the rclone connection string.

task/task.go
```diff
@@ -49,11 +49,9 @@ func New(ctx context.Context, cloud common.Cloud, identifier common.Identifier,
 	}
 }
 
+// Task defines the interface implemented by provided-specific task resources.
```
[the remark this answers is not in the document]
tasdomas: Thank you!

task/common/resource.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/651/commits/07e4116a3ff87fdb6ba69aeb224b723fa1e5840f

task/common/resource.go
0x2b3bfa0: > _So, using cloud credentials to retrieve azure storage credentials will only work if the storage is owned by the same azure account._

What do you mean by account? 🤔

task/aws/resources/data_source_bucket_test.go
0x2b3bfa0: If your proposal implies accounting for a considerable amount of cloud-specific details (e.g. “limited interfaces” for S3 and EC2, others for Cloud Storage and Compute Engine, ...) I wonder if maintaining them will be any easier than testing the code against actual cloud APIs, even if it's noticeably slower.

task/common/values.go
tasdomas: what about e.g. a minio bucket?

0x2b3bfa0: That's not a native storage strictly speaking, and would have the same consideration as a Google Cloud task using a Microsoft Azure storage account. It would be out of scope for us, and users would be responsible of synchronization as part of their code.

task/common/resource.go
tasdomas: By `account` I meant the subscription - sorry about not being clear enough.

tasdomas: > https://github.com/iterative/terraform-provider-iterative/blob/66e18e4dcacc798b7d9a33a686fb991be4c37eda/task/az/resources/resource_storage_account.go#L70
This only works if the storage account is in the same resource group. In our case (since we've just created the resource group) we'd need to pass the resource group that the already existing storage account belongs to.

0x2b3bfa0: It seems to me a reasonable requirement. If the storage account were in a different subscription, it would have to be treated as an user-controlled storage (e.g. DVC remote) and not as a task storage.

I'd prefer to keep task storage as local and native as possible, and leave other kinds of data synchronization to specialized tools.

tasdomas: ok, this still leaves the issue of the storage account and its access key

tasdomas: > > _the issue of the storage account and its access key_
> 
> question
Well, in order to access a blob container, you need a storage account. And to retrieve a storage account's key you need to know its resource group, so either way the user needs to supply either `(storage_account, key)` or `(resource_group, storage_account)`.

0x2b3bfa0 (approved):
None of the review comments are specific to AWS or GCP. Feel free to merge into the feature branch. 🌳

PR #653, opened 2022-09-03
Implement human-readable random identifier generation

0x2b3bfa0:
Now users can run e.g. `task create --name experiment` to create a task named `experiment` with a randomly generated suffix, avoiding clashes due to cloud providers' globally unique resource name requirements.

> **Note**
> Terraform resources' behavior remains unaltered and fully backwards-compatible. This change only affects the standalone command-line tool.

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -102,13 +102,9 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
 		cfg.Spot = common.Spot(common.SpotEnabled)
 	}
 
-	id := common.NewRandomIdentifier()
-
-	if o.Name != "" {
-		id = common.NewIdentifier(o.Name)
-		if identifier, err := common.ParseIdentifier(o.Name); err == nil {
-			id = identifier
-		}
+	id := common.NewRandomIdentifier(o.Name)
+	if identifier, err := common.ParseIdentifier(o.Name); err == nil {
+		id = identifier
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), cloud.Timeouts.Create)
--- a/task/common/identifier.go
+++ b/task/common/identifier.go
@@ -68,9 +68,9 @@ func (i Identifier) Short() string {
 }
 
 // hash deterministically generates a Base36 digest of `size`
-// characters using the provided salt.
-func hash(salt string, size uint8) string {
-	digest := sha256.Sum256([]byte(salt))
+// characters using the provided seed.
+func hash(seed string, size uint8) string {
+	digest := sha256.Sum256([]byte(seed))
 	random := uid.NewRandCustom(bytes.NewReader(digest[:]))
 	encoder := uid.NewEncoderBase36()
 	provider := uid.NewProviderCustom(sha256.Size, random, encoder)
--- a/task/common/identifier.go
+++ b/task/common/identifier.go
@@ -10,6 +10,7 @@ import (
 	"crypto/sha256"
 
 	"github.com/aohorodnyk/uid"
+	"github.com/dustinkirkland/golang-petname"
 )
 
 type Identifier struct {
@@ -49,7 +50,8 @@ func NewDeterministicIdentifier(name string) Identifier {
 func NewRandomIdentifier(name string) Identifier {
 	seed := uid.NewProvider36Size(8).MustGenerate().String()
 	if name == "" {
-		name = seed
+		petname.NonDeterministicMode()
+		name = petname.Generate(3, "-")
 	}
 
 	return Identifier{name: name, salt: hash(seed, shortLength/2)}
--- a/task/common/identifier.go
+++ b/task/common/identifier.go
@@ -35,11 +35,17 @@ func ParseIdentifier(identifier string) (Identifier, error) {
 	return Identifier{}, ErrWrongIdentifier
 }
 
+// NewDeterministicIdentifier returns a new deterministic Identifier, using the
+// provided name as a seed. Repeated calls to this function are guaranteed to
+// generate the same Identifier.
 func NewDeterministicIdentifier(name string) Identifier {
 	seed := normalize(name, nameLength)
 	return Identifier{name: name, salt: hash(seed, shortLength/2)}
 }
 
+// NewRandomIdentifier returns a new random Identifier. Repeated calls to this
+// function are guaranteed to generate different Identifiers, as long as there
+// are no collisions.
 func NewRandomIdentifier(name string) Identifier {
 	seed := uid.NewProvider36Size(8).MustGenerate().String()
 	if name == "" {
--- a/task/common/identifier.go
+++ b/task/common/identifier.go
@@ -12,38 +12,46 @@ import (
 	"github.com/aohorodnyk/uid"
 )
 
-type Identifier string
+type Identifier struct {
+	name string
+	salt string
+}
 
 var ErrWrongIdentifier = errors.New("wrong identifier")
 
 const (
 	maximumLongLength = 50
 	shortLength       = 16
+	nameLength        = maximumLongLength-shortLength-uint32(len("tpi---"))
 )
 
 func ParseIdentifier(identifier string) (Identifier, error) {
 	re := regexp.MustCompile(`(?s)^tpi-([a-z0-9]+(?:[a-z0-9-]*[a-z0-9])?)-([a-z0-9]+)-([a-z0-9]+)$`)
 
 	if match := re.FindStringSubmatch(string(identifier)); len(match) > 0 && hash(match[1]+match[2], shortLength/2) == match[3] {
-		return Identifier(match[1]), nil
+		return Identifier{name: match[1], salt: match[2]}, nil
 	}
 
-	return Identifier(""), ErrWrongIdentifier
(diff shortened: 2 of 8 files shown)


cmd/create/create.go
```diff
@@ -102,13 +102,9 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
...
 
-	id := common.NewRandomIdentifier()
-
-	if o.Name != "" {
-		id = common.NewIdentifier(o.Name)
-		if identifier, err := common.ParseIdentifier(o.Name); err == nil {
-			id = identifier
-		}
+	id := common.NewRandomIdentifier(o.Name)
+	if identifier, err := common.ParseIdentifier(o.Name); err == nil {
```
tasdomas, line 106: A tangential question - why is the identifier format (and its parsing) so complicated:
```
	re := regexp.MustCompile(`(?s)^tpi-([a-z0-9]+(?:[a-z0-9-]*[a-z0-9])?)-([a-z0-9]+)-([a-z0-9]+)$`)

	if match := re.FindStringSubmatch(string(identifier)); len(match) > 0 && hash(match[1]+match[2], shortLength/2) == match[3] {
```

task/common/identifier.go
```diff
@@ -12,38 +12,46 @@ import (
...
+	name string
+	salt string
+}
 
 var ErrWrongIdentifier = errors.New("wrong identifier")
 
 const (
 	maximumLongLength = 50
 	shortLength       = 16
+	nameLength        = maximumLongLength-shortLength-uint32(len("tpi---"))
```
tasdomas, line 25: why `tpi---` ?

task/common/identifier.go
```diff
@@ -12,38 +12,46 @@ import (
...
+		return Identifier{name: match[1], salt: match[2]}, nil
 	}
 
-	return Identifier(""), ErrWrongIdentifier
+	return Identifier{}, ErrWrongIdentifier
 }
 
-func NewIdentifier(identifier string) Identifier {
-	return Identifier(identifier)
+func NewDeterministicIdentifier(name string) Identifier {
```
tasdomas, line 38: Public functions need a comment describing what they do.

task/common/identifier.go
```diff
@@ -12,38 +12,46 @@ import (
...
+type Identifier struct {
+	name string
+	salt string
+}
 
 var ErrWrongIdentifier = errors.New("wrong identifier")
 
 const (
 	maximumLongLength = 50
 	shortLength       = 16
```
tasdomas, line 24: I find it curious that in most places in the code we don't refer to `shortLength`, but rather `shortLength/2`

tasdomas (changes requested):
Generally looks good, but..

exposed declarations need a short docstring identifying them, for an overview of go documentation guidelines see [go.dev](https://go.dev/doc/comment)

task/common/identifier.go
0x2b3bfa0: * Long identifiers are in the form `tpi-{1}-{2}-{3}` and are menat to be used for most cloud resources.
* Short identifiers are in the form `{2}{3}` and are meant to be used for a few cloud resources with severe name length constraints.

The `shortLength` constant specifies the total length of a short identifier, which is equal to the sum of the lengths for `{2}` and `{3}`. Thus, `{2}` and `{3}` are being generated with a legth limit of `shortLength/2`.

0x2b3bfa0: * `{1}` is a normalized human-friendly name, generated from the `--name` parameter[^1] when provided, or randomly when omitted.
* `{2}` is either a random string (for random identifiers) or a deterministic string derived from `{1}` (for deterministic identifiers)
* `{3}` is just a hash of `{1}` and `{2}` so we can tell apart the cloud resources we created from any other resources


[^1]: e.g. if you run `task create --name=Example@Name` the resulting `{1}` will be `example-name`

task/common/identifier.go
0x2b3bfa0: `len("tpi---")` is the length of the `tpi` prefix plus the three hyphens separating fields `{1}`, `{2}` and `{3}`. I thought it would be less confusing than a hardcoded `6` but probably it isn't the case. 🤷🏼‍♂️

cmd/create/create.go
0x2b3bfa0: See below

task/common/identifier.go
tasdomas: So `{3}` is just a checksum portion to make it less likely that somebody will create a resource name matching `tpi`'s conventions?

task/common/identifier.go
tasdomas: This should be in a comment above `nameLength` - the perfect example of explaining _why_ :)

I'd probably just define `maximumLength` as a combination of `nameLength` and `shortLength`.

task/common/identifier.go
0x2b3bfa0: Yes, `{3}` is just a small hash to tell apart any resource named with a `tpi-` (or `leo-` soon) prefix from resources we create.

task/common/identifier.go
tasdomas: Thanks!

Not sure what you mean by "has no public API" - if a package has exported identifiers, it has what can be considered a public API.

task/common/identifier.go
```diff
@@ -12,38 +12,52 @@ import (
 	"github.com/aohorodnyk/uid"
 )
 
-type Identifier string
+type Identifier struct {
```
tasdomas, line 15: Ah, this could use a comment too.

tasdomas (approved):
lgtm

task/common/identifier.go
0x2b3bfa0: I mean that indeed, it has a “public API” in the sense that we can consume it ourselves, but the module is a Terraform provider and, as such:

1. Can't be imported by other projects.
2. Doesn't even appear in pkg.go.dev

task/common/identifier.go
0x2b3bfa0: > _This should be in a comment above nameLength - the perfect example of explaining why_

Comments, comments, ... 😈 

> _I'd probably just define maximumLength as a combination of nameLength and shortLength._

Sounds good, although I'm a bit reluctant to do that. I wanted the total length limits to be really obvious, as they're the important numbers. E.g. resource X names on cloud Y can have a maximum length of Z characters, and our names are below that number.

tasdomas: I see where you're coming from, but still - `maximumLength` is a function of `(shortLength, nameLength)`, not the other way around. And if we want to restrict `maximumLength` from exceeding a specific value, go's tests are a great way to do that.

task/common/identifier.go
tasdomas: Surely we're not documenting things just for the sake of third-party developers, right?

The fact that we make tpi non-trivial to import into another go project by misdefining the module name has no bearing. And I'm not sure what you mean by saying it can't be imported into other projects - I whip up a working example of doing exactly that in a few minutes.

0x2b3bfa0: > _Surely we're not documenting things just for the sake of third-party developers, right?_

For instance, DVC only has “docstrings” on those files/classes/functions that are meant to be imported by end users.  🤭 But yes, we should be the main audience for code documentation if we choose to.

0x2b3bfa0: > _The fact that we make tpi non-trivial to import into another go project by misdefining the module name has no bearing._

It does, in the sense that whatever we document won't be useful for anybody else than developers and contributors.

0x2b3bfa0: I guess that this conversation boils down to whether we want comments for ourselves or not. 🤔

0x2b3bfa0: Resolving in favor of notion#LEO-improvements-a763b3a3308749e68ad9b53cc1438e8c

task/common/identifier.go
0x2b3bfa0: > _`maximumLength` is a function of `(shortLength, nameLength)`, not the other way around_

Why can't it be the other way around? 🤔 If you want to avoid operations other than addition, you're right, but... would it be any easier to reason about in that way?





> _And if we want to restrict `maximumLength` from exceeding a specific value, go's tests are a great way to do that._

Isn't that a bit more convoluted than the current approach? 🤔

tasdomas:
Let's pass the tests and land this - contrary to popular belief PRs do not need time to "breathe" :)

cmd/create/create.go
tasdomas: Thanks for the explanation - much clearer now.

tasdomas (approved):

0x2b3bfa0:
> _contrary to popular belief PRs do not need time to "breathe"_

Pull requests probably don't, but developers do. 🤣

0x2b3bfa0:
Let me see if I can pass the tests: this pull request modifies code that is common to every “cloud” and I would like all the smoke tests to pass.

task/common/identifier.go
0x2b3bfa0: I'm afraid I'll leave it for the next time. 😅

0x2b3bfa0:
It looks like Kubernetes succeeded, despite the last failed run; merging...

PR #654, opened 2022-09-04
Print identifier before creating resources

0x2b3bfa0:
Suggested by @tasdomas, simplifies manual cleanup if the process fails


0x2b3bfa0:
* Merged into #655

PR #655, opened 2022-09-05
Simplify command-line tool output format

0x2b3bfa0:


cmd/read/read.go
```diff
@@ -30,6 +32,10 @@ func New(cloud *common.Cloud) *cobra.Command {
 	}
 
 	cmd.Flags().IntVar(&o.Parallelism, "parallelism", 1, "parallelism")
+	cmd.Flags().BoolVar(&o.Status, "status", false, "Read status")
+	cmd.Flags().BoolVar(&o.Events, "events", false, "Read events")
```
tasdomas, line 36: What's the difference between events and logs?

cmd/read/read.go
```diff
@@ -58,45 +64,50 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
...
-	for _, event := range tsk.Events(ctx) {
-		events = append(events, fmt.Sprintf(
-			"%s: %s\n%s",
-			event.Time.Format("2006-01-02 15:04:05"),
-			event.Code,
-			strings.Join(event.Description, "\n"),
-		))
-	}
-	logrus.Info(events)
+	switch {
```
tasdomas, line 67: Shouldn't this be just three conditions instead of a `switch`? Go's `switch` is not _fallthrough_ by default.

cmd/read/read.go
0x2b3bfa0: Events are “scaling events” like “requested 3 instances, there is not enough capacity to satisfy the request”; logs are the output of the `script` provided by the user.

cmd/read/read.go
0x2b3bfa0: I guess it doesn't make any difference for this use case, because `o.Status`, `o.Events` and `o.Logs` are mutually exclusive. 🤷🏼‍♂️ 

Although separate conditions are probably cleaner, or even `switch` and separate functions for each `case`; cramming all the code there is probably a bit too much.

tasdomas: Having three separate flags define something that's exclusive is counter-intuitive. A nice principle to follow here is to not make it possible to specify a flag combination that is invalid (i.e. `-status -events`). I'd probably go with a `-show [events|logs|status]` flag then.

0x2b3bfa0: Using [`MarkFlagsMutuallyExclusive`](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) doesn't seem precisely counter-intuitive to my mind. See for instance `gh pr merge [--merge|--rebase|--squash]` where `only one of --merge, --rebase, or --squash can be enabled` likewise.

0x2b3bfa0: Although, admittedly, options on `gh pr merge` represent verbs and options on `task read` represent nouns. 🤔

tasdomas: I missed [MarkFlagsMutuallyExclusive](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) - I think it's fine as it is

cmd/read/read.go
0x2b3bfa0: Moved events to `DEBUG` level logs in `--status`

cmd/read/read.go
0x2b3bfa0: Preoposed solution on https://github.com/iterative/terraform-provider-iterative/pull/655/commits/b7722fd70d13b883bbe6d8f900804b6f7a740e3f

cmd/read/read.go
omesser: @0x2b3bfa0 
so, may I suggest:
`o.Events` -> `o.ScalingEvents` or `o.InfrastructureEvents`?

0x2b3bfa0: `o.Events` doesn't exist anymore, as it was “merged” with `o.Status` 🤔

0x2b3bfa0: Would you suggest adding again a third option (e.g. `--scaling-events`) to print those events to standard output?

cmd/create/create.go
```diff
@@ -115,6 +116,9 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
 		return err
 	}
 
+	logrus.Infof("Using identifier %s", id.Long())
+	defer fmt.Println(id.Long())
```
tasdomas, line 120: Do we need this `defer` here?

cmd/root.go
```diff
@@ -61,18 +61,21 @@ func New() *cobra.Command {
 	cmd.AddCommand(stop.New(&o.Cloud))
 
 	cmd.PersistentFlags().StringVar(&o.Provider, "cloud", "", "cloud provider")
-	cmd.PersistentFlags().StringVar(&o.Log, "log", "info", "log level")
+	cmd.PersistentFlags().CountVar(&o.Verbose, "verbose", "verbose output")
```
tasdomas, line 64: Why `CountVar` and not just `BoolVar`?

tasdomas (approved):
Generally looks good, just a few questions

cmd/create/create.go
0x2b3bfa0: It was just to make sure that the identifier is printed to `stdout` at the end, both in case of error and success. Seeing it on the top of the logs wasn't particularly pleasant for me.

My goal was:
1. Print the identifier to the **`stderr`** user-oriented logs in the beginning.
2. Print the identifier to the **`stdout`** unix-oriented logs at the end.

0x2b3bfa0: If it's too much, we can just **(2)** in the beginning as you originally proposed.

cmd/root.go
0x2b3bfa0: Just in case we introduced additional debug levels and a `-v` shorthand, so we can act differently upon `-v`, `-vv`, `-vvv`, ...

0x2b3bfa0: We can use a `BoolVar` until that functionality is really needed, though. It was probably a premature design choice.

tasdomas: I think we can should make it a `BoolVar` for now.

cmd/create/create.go
tasdomas: I see. Ok, let's keep it as is for now.

cmd/root.go
0x2b3bfa0: Resolved with  304c496

tasdomas (approved):

0x2b3bfa0:
Merging with some smoke tests failing because of unrelated resource availability issues

PR #661, opened 2022-09-08
K8s support for specifying an existing persistent volume claim

tasdomas:


task/k8s/resources/data_source_persistent_volume.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// VolumeInfo returns information for attaching the persistent volume claim to the job.
+func (p *ExistingPersistentVolumeClaim) VolumeInfo(ctx context.Context) (string /*subpath*/, *kubernetes_core.PersistentVolumeClaimVolumeSource) {
+	pvc := &kubernetes_core.PersistentVolumeClaimVolumeSource{
+		ClaimName: p.params.Container,
+	}
+	return p.params.Path, pvc
+
+}
```
0x2b3bfa0, line 52: Why use a function plus an interface instead of simply exposing the values as part of the `ExistingPersistentVolumeClaim` structure? 🤔

tasdomas: Because we have two implementations for that interface - `PersistentVolumeClaim` and `ExistingPersistentVolumeClaim`

0x2b3bfa0: Why not just use a separate data source like in the other backends? 🤔

tasdomas: It is a data source. This is the exact same pattern used in other backends. We're just using the interface to pass to the `job` resource.

tasdomas:
This can be tested with the following terraform config (`main.tf`):
```
terraform {
}

provider "kubernetes" {
  config_path    = "~/.kube/config"
  config_context = "minikube"
}

# A pre-allocated PVC.
resource "kubernetes_persistent_volume_claim" "example" {
  metadata {
    name = "pvc"
  }
  spec {
    access_modes = ["ReadWriteMany"]
    resources {
      requests = {
        storage = "1Gi"
      }
    }
  }
}

# A busybox pod to let us inspect the contents of the PVC.
resource "kubernetes_pod" "pvsee" {
  metadata {
    name = "pvsee"
  }

  spec {
    container {
      image = "busybox"
      name  = "bb"
      command = ["tail", "-f", "/dev/null"]
      volume_mount {
        name = "data"
        mount_path = "/data"
      }
    }
    volume {
      name = "data"
      persistent_volume_claim {
        claim_name = "${kubernetes_persistent_volume_claim.example.metadata.0.name}"
      }
    }
  }
}

provider "iterative" {}

  cloud      = "k8s" # or any of: gcp, az, k8s
  region     = "us-east"
  machine    = "m"   # medium. Or any of: l, xl, m+k80, xl+v100, ...
  spot       = -1     # auto-price. Default -1 to disable, or >0 for hourly USD limit
  disk_size  = -1    # GB. Default -1 for automatic
  storage {
    workdir = "./"       # default blank (don't upload)
    output  = "results" # default blank (don't download). Relative to workdir
    container = "${kubernetes_persistent_volume_claim.example.metadata.0.name}"
    container_path = "tpi-run"
  }
  script = <<-END
    #!/bin/bash

    # create output directory if needed
    mkdir -p results
    # read last result (in case of spot/preemptible instance recovery)
    if test -f results/epoch.txt; then EPOCH="$(cat results/epoch.txt)"; fi
    EPOCH=$${EPOCH:-1}  # start from 1 if last result not found

    echo "(re)starting training loop from $EPOCH up to 1337 epochs"
    for epoch in $(seq $EPOCH 1337); do
      sleep 1
      echo "$epoch" | tee results/epoch.txt
    done
  END
}

```

Once the config is applied, run `kubectl exec -it pvsee -- sh` and then `ls /data` to inspect the contents of the PVC.

0x2b3bfa0 (approved):
Sorry for letting this pull request [“breathe”](https://github.com/iterative/terraform-provider-iterative/pull/653#issuecomment-1241587241) for too long. 🙈 I believe it'll be more practical to merge it and submit a global review once you open a pull request for the feature branch-

PR #662, opened 2022-09-09
Separate leo and tpi executables

tasdomas:


0x2b3bfa0 (changes requested):
To my mind, moving the Terraform Provider entry point under `./cmd/tpi` makes no sense from a project structure standpoint. Note that this repository is first and foremost a Terraform Provider.

I'd wait to refine `leo` user experience a bit more and then suggest migrating it to a different module, so it can be easily installed with `go install` like any other tool, and the `iterative_task` Terraform Resource can still depend on the underlying packages.

tasdomas:
> To my mind, moving the Terraform Provider entry point under `./cmd/tpi` makes no sense from a project structure standpoint. Note that this repository is first and foremost a Terraform Provider.
> 
> I'd wait to refine `leo` user experience a bit more and then suggest migrating it to a different module, so it can be easily installed with `go install` like any other tool, and the `iterative_task` Terraform Resource can still depend on the underlying packages.

Ok, I've moved `tpi` back into project root.

.goreleaser.yml
```diff
@@ -27,14 +27,43 @@ builds:
       goarch: arm64
     - goos: windows
       goarch: arm
-  binary: '{{ .ProjectName }}_v{{ .Version }}'
+  binary: 'terraform-provider-iterative'
+-
+  id: "leo"
+  main: ./cmd/leo
```
0x2b3bfa0, line 33: As well, moving files from `./cmd` to `./cmd/leo` seems a bit redundant to me. 🤔 How many other binaries are we planning to generate from this repository? 😅

task/common/machine/testdata/machine_script_minimal.golden
```diff
@@ -62,9 +62,10 @@ sudo tee /etc/systemd/system/tpi-task.service > /dev/null <<END
 END
 
 curl --location --remote-name https://github.com/iterative/terraform-provider-iterative/releases/latest/download/terraform-provider-iterative_linux_amd64
-sudo mv terraform-provider-iterative* /usr/bin/tpi
-sudo chmod u=rwx,g=rx,o=rx /usr/bin/tpi
-sudo chown root:root /usr/bin/tpi
+# TODO: replace download location with https://github.com/iterative/terraform-provider-iterative/releases/latest/download/leo_linux_amd64
+sudo mv leo* /usr/bin/leo
```
0x2b3bfa0, line 66:
```suggestion
sudo mv terraform-provider-iterative* /usr/bin/leo
```

.goreleaser.yml
tasdomas: By convention `cmd` is usually a container package for other commands. I don't think it hurts being explicit about command groupings.

task/common/machine/testdata/machine_script_minimal.golden
tasdomas: Thanks!

0x2b3bfa0:
As stated at https://github.com/iterative/terraform-provider-iterative/pull/662#pullrequestreview-1102257207, prematurely separating binaries in the same repository might not add too much value, and it's probably more important to separate the repositories themselves. 🤔

tasdomas:
I'd argue that the binaries were combined prematurely. And that is causing a [bug](https://github.com/iterative/terraform-provider-iterative/issues/640).

0x2b3bfa0:
I'd argue[^1] that whether binaries are combined or not is not an issue, but trying to shoehorn Terraform HCL on a Cobra & Viper command-line tool was definitely premature (or, rather, something that shouldn't have happened)

[^1]: Although agreeing is a strange way of arguing

0x2b3bfa0:
I don't want to stop this pull request for the sake of perfectionism; it's indeed a step in the right direction, degree more or less 🌏 🌍 🌎

0x2b3bfa0 (approved):

PR #664, opened 2022-09-09
Some refactoring on leo subcommands

tasdomas:
Renamed delete to destroy.
Renamed read to status.
Restructured base options and moved all subcommands into a single package.


0x2b3bfa0 (commented):
> _moved all subcommands into a single package_

Why? 🤔

tasdomas:
> > _moved all subcommands into a single package_
> 
> Why? thinking

Is there any reason to keep identical single file packages?

0x2b3bfa0:
> _Is there any reason to keep identical single file packages?_

~“Namespaces are one honking great idea — let's do more of those!”~ Wrong language! 🤭 

It probably boils down to a matter of taste[^1] but, to my mind, it looks cleaner once we introduce subcommands or more complex commands (e.g. #291 or #531) which may be split over several files.

See also [docker/cli](https://github.com/docker/cli), [kubernetes/kubectl](https://github.com/kubernetes/kubectl) or [cli/cli](https://github.com/cli/cli) for inspiration.

[^1]: Or lack thereof, you're the wise gopher here.

cmd/root.go
```diff
@@ -146,10 +108,10 @@ func New() *cobra.Command {
 		}
 	}
 
-	for _, cmd := range append(cmd.Commands(), cmd) {
+	for _, subcmd := range append(cmd.Commands(), cmd) {
```
0x2b3bfa0, line 111: `-Winaccurate-identifier` — we aren't just iterating over subcommands, but also over the main command; glad to find any other name to avoid a clash, but `subcmd` is slightly misleading.

tasdomas: Resolved with [ff9c6a0](https://github.com/iterative/terraform-provider-iterative/pull/664/commits/ff9c6a040bd02e35f61c6e34fbce855aa8b59a40)

0x2b3bfa0 (commented):
I'd still prefer to keep subcommands in separate packages, unless there is a good reason not to.

_Sample good reason: the structure you propose eliminates beautifully the `common.Cloud` “dependency injection” used by the previous approach._

<details><summary>Relevant XKCD</summary>

[I'm not sure if this is actually true](https://xkcd.com/106)

</details>

cmd/common.go
```diff
@@ -0,0 +1,63 @@
...
+	logrus.SetLevel(logrus.InfoLevel)
+	if o.Verbose {
+		logrus.SetLevel(logrus.DebugLevel)
+	}
+
+	logrus.SetFormatter(&logrus.TextFormatter{
+		ForceColors:      true,
+		DisableTimestamp: true,
+	})
+}
```
0x2b3bfa0, line 58: Message in a botte for myself: lots of good stuff here, don't let them disappear forever

Issue #648, opened 2022-08-29
Avoid using environment variables to alter Kubernetes task functionality


0x2b3bfa0:
#### Extracted from https://github.com/iterative/terraform-provider-iterative/pull/645#discussion_r952928867

Kubernetes `task` code uses environment variables in a rather hacky way to differentiate file transfer from code execution jobs. While this is mostly a consequence of the approach we follow, it would be nice to use function arguments or something more explicit.

Issue #647, opened 2022-08-29
Kubernetes task deletion logic is flawed


0x2b3bfa0:
It's either a bug or some weird workaround @0x2b3bfa0 forgot to document. Sounds like a bug.

https://github.com/iterative/terraform-provider-iterative/blob/7b27c579a56813c6f48627fe6da777d63f46808b/task/k8s/task.go#L205-L207

Should read `t.Resources.Job.Delete` instead of `t.Resources.Job.Create` but changing it breaks the tests.

PR #669, opened 2022-09-14
Fix k8s data transfer setup

tasdomas:
Instead of reusing (abusing) the original job, launch a separate deployment with a busybox pod and minimal requirements to facilitate data transfer.

This addresses #647 and #648


task/k8s/resources/resource_tx_pod.go
```diff
@@ -0,0 +1,146 @@
+package resources
```
0x2b3bfa0, line 1: Consider renaming `resource_tx_pod.go` to `resource_transfer_pod.go` so it can't be mistaken with _transaction_ nor _transmission_. Blockchain and embedded system developers will be extremely thankful. 😄

task/k8s/resources/resource_tx_pod.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Create creates the transfer pod resource.
+func (p *TransferPod) Create(ctx context.Context) error {
+	if p.Attributes.Task.Environment.Directory == "" {
+		return fmt.Errorf("output directory not set")
+	}
+
+	volumeMounts := []k8s_core.VolumeMount{{
+		Name:      p.Identifier + "-pvc",
```
0x2b3bfa0, line 52: Suffixes like `-pvc` are rather futile, given that Kubernetes resources are already classified into different collections. E.g. there won't be a name clash between a Pod named `example` and a namesake PersistentVolume.

task/k8s/resources/resource_tx_pod.go
```diff
@@ -0,0 +1,146 @@
...
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	k8s_meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+
+	"terraform-provider-iterative/task/common"
+	"terraform-provider-iterative/task/k8s/client"
+)
+
+// NewTransferPod creates a pod used to transfer data between the persistent storage and local.
+func NewTransferPod(client *client.Client, identifier common.Identifier, persistentVolumeClaim *PersistentVolumeClaim, permissionSet *PermissionSet, task common.Task) *TransferPod {
```
0x2b3bfa0, line 19: Would `TransferDeployment` be more adequate? 🤔

task/k8s/resources/resource_tx_pod.go
```diff
@@ -0,0 +1,146 @@
...
+	return p
+}
+
+// TransferPod is a deployment with a single pod used to transfer data between the persistent volume claim and local.
+type TransferPod struct {
+	Client     *client.Client
+	Identifier string
+	Attributes struct {
+		Task      common.Task
+		Addresses []net.IP
```
0x2b3bfa0, line 36: What do we use this for? 🤔

0x2b3bfa0 (commented):
This code does not guarantee that the Deployment is going to be scheduled to the same node as the subsequent Job, effectively invalidating the use of access modes other than [ReadWriteMany](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes). At least, sounds worth documenting, methinks.

Would there be any more succint way of doing this before having to maintain a separate Spec for data transfer?

task/k8s/resources/resource_tx_pod.go
tasdomas: True. Thanks!

task/k8s/resources/resource_tx_pod.go
tasdomas: It would. Updating.

task/k8s/resources/resource_tx_pod.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/669/commits/97cb39e774d9d00790402afeaa544fd19f565a37

task/k8s/resources/resource_tx_pod.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/669/commits/97cb39e774d9d00790402afeaa544fd19f565a37

task/k8s/resources/resource_tx_pod.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/669/commits/97cb39e774d9d00790402afeaa544fd19f565a37

task/k8s/resources/resource_tx_pod.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/669/commits/97cb39e774d9d00790402afeaa544fd19f565a37

0x2b3bfa0:
Instead of separating data transfer from the main job, we should consider using a [sidecar](https://k8spatterns.io/) or, probably better yet, an [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) as part of the main job, with the sole purpose of performing data transfer. What do you think?

tasdomas:
Won't the init container be started for each job pod though?

0x2b3bfa0:
Yes, although that's a feature rather than a bug. 😄 We can use the init container for synchronization when parallelism is greater than 1.

E.g. use the first init container for the actual data synchronization process, and use the others just to wait until the data copy finishes.

0x2b3bfa0:
There's still an issue, though: copying the results back once the job finishes running.

PR #671, opened 2022-09-15
Mechanical cleanups and style fixes

tasdomas:
Do not use anonymous struct members.
Make client a private member.
Use literals instead of `new(type)` where possible.

### DIFF
--- a/task/aws/client/client.go
+++ b/task/aws/client/client.go
@@ -34,13 +34,12 @@ func New(ctx context.Context, cloud common.Cloud, tags map[string]string) (*Clie
 		return nil, err
 	}
 
-	c := new(Client)
-	c.Cloud = cloud
-	c.Region = region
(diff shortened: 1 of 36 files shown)


0x2b3bfa0 (commented):
> _Use literals instead of `new(type)` where possible_

In most of the project's code, “where possible” doesn't equate “everywhere” and, for the sake of consistency, I chose to use the also widespread `new(type)` pattern uniformly.

Still, you proposed change is “dr[y]er” and looks readable to me.

0x2b3bfa0 (changes requested):
Looks mergeable to me, but all the tests (except for the `k8s` ones) are timing out. Have you perhaps unwittingly introduced some modification to the log retrieval logic? 🤔

0x2b3bfa0 (approved):
It was just some Deity of the Flaky Tests being playful

PR #443, opened 2022-03-14
docs: permissions & reference roles

0x2b3bfa0:
Closes #75; may need extra eyes and opinions on documentation and some awkward `kubectl` scripting.

### Potential follow–ups
* [ ] Validate all this code as part of our CI/CD `smoke` test workflow[^1]
* [ ] Migrate Kubernetes YAML to HCL for consistency[^2]
* [ ] Provide tighter ABAC apart from basic RBAC permissions[^3]

> ### Exporting credentials (one–liner)
> ```console
> eval "$(terraform output --json | jq --raw-output 'to_entries[]|"export \(.key|ascii_upcase)=\(.value.value|@sh)"')"
> ```
> Note: added `eval` to support multi–line variables like `gcp` service account credentials.
> 
> _Originally posted by @0x2b3bfa0 in https://github.com/iterative/terraform-provider-iterative/issues/75#issuecomment-1057074357_

[^1]: Requires some internal infrastructure work to create the appropriate sandboxes.
[^2]: Doesn't seem possible without a https://github.com/hashicorp/terraform-provider-kubernetes/issues/1641
[^3]: https://twitter.com/michaelgat/status/1017166804139954176

PR #673, opened 2022-09-16
Make smoke tests more lightweight

tasdomas:
Currently smoke tests call tpi to provision machines with GPUs. This often runs into resource quotas.
The proposal runs PR tests provisioning smaller (m) machines and a test to provision machines with GPUS is 
run on a fixed daily schedule on master.


task/task_smoke_test.go
```diff
@@ -19,19 +19,62 @@ import (
...
+#!/bin/sh -e
+nvidia-smi
+mkdir --parents cache output
+touch cache/file
+echo "$ENVIRONMENT_VARIABLE_DATA" | tee --append output/file
+sleep 60
+cat output/file
+`[1:],
+	},
+}
```
0x2b3bfa0, line 55: I don't quite like the idea of maintaining two copies of the script just to skip a line. Instead of parameterizing `script` you could parameterize `environment` and set `TEST_GPU=true` so the script can branch accordingly.[^1]

[^1]: Depending on how purist you are, `$TEST_GPU && nvidia-smi` or `printenv TEST_GPU && nvidia-smi` can be chosen over `test -z "$TEST_GPU" && nvidia-smi` or a dozen other combinations.

.github/workflows/gpu-smoke-test.yml
```diff
@@ -0,0 +1,113 @@
...
+        KUBECONFIG_DATA: ${{ steps.cluster.outputs.kubeconfig }}
+        SMOKE_TEST_ENABLE_K8S: true
+        SMOKE_TEST_TARGET: gpu
+    - if: always()
+      run: |
+        az group delete \
+          --name="tpiSmokeTestCluster$GITHUB_RUN_ID" \
+          --yes
+env:
+  SMOKE_TEST_IDENTIFIER: smoke test ${{ github.run_id }}
```
0x2b3bfa0, line 114: In continuous integration workflows, a little dependency is much worse than a ~little~ huge copying. Consider using [`if`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif) or other gating mechanisms to reuse the existing workflow.

0x2b3bfa0 (changes requested):

0x2b3bfa0 (commented):
Stretch goal: make also spot instance testing optional

https://github.com/iterative/terraform-provider-iterative/blob/970711e251ebb68d4ae9617724a4ef937b62c8d0/task/task_smoke_test.go#L103

task/task_smoke_test.go
```diff
@@ -19,19 +19,62 @@ import (
...
 	enableALL := !enableAWS && !enableAZ && !enableGCP && !enableK8S
 
+	testTargetName := os.Getenv("SMOKE_TEST_TARGET")
+	if testTargetName == "" {
+		testTargetName = "default"
+	}
+	target, ok := testTargets[testTargetName]
+	if !ok {
+		t.Fatalf("test target %q undefined", testTargetName)
+	}
```
0x2b3bfa0, line 76: Nitpick: maybe it's enough to use `targetName` for that identifier? 💅🏼 
```suggestion
	targetName := os.Getenv("SMOKE_TEST_TARGET")
	if targetName == "" {
		targetName = "default"
	}
	target, ok := testTargets[targetName]
	if !ok {
		t.Fatalf("test target %q undefined", targetName)
	}
```

task/task_smoke_test.go
tasdomas: ok, updated

.github/workflows/gpu-smoke-test.yml
tasdomas: updated

.github/workflows/smoke.yml
```diff
@@ -5,7 +5,8 @@ on:
   pull_request_target:
     paths: 'task/**'
   schedule:
-  - cron: '0 9 * * *'
+  - cron: '0 1 * * *'
```
dacbd, line 8:
```suggestion
  - cron: '0 1 * * *' # Everyday @ 0100 UTC
```

dacbd (approved):
lgtm

task/task_smoke_test.go
```diff
@@ -19,19 +19,51 @@ import (
...
+		spot:    common.SpotDisabled,
+	},
+	"gpu": testTarget{
+		machine: "m+t4",
+		image:   "nvidia",
+		env: map[string]string{
+			"TEST_GPU": "yes",
+		},
+		spot: common.SpotEnabled,
+	},
```
0x2b3bfa0, line 43: Cramming the GPU and spot tests under the `gpu` target migh be a bit too much.

I'd either:
1. Rename targets to e.g. “short” and “long” or “fast” and “slow” or “partial” and “full” or “consistent” and “flaky” ...
2. Create separate targets for both `gpu` and `spot` options and make them somehow composable.

task/task_smoke_test.go
```diff
@@ -42,7 +74,17 @@ func TestTaskSmoke(t *testing.T) {
...
+#!/bin/sh -e
+if [ -n "$TEST_GPU" ]; then
+  nvidia-smi
+fi
+mkdir --parents cache output
+touch cache/file
+echo "$ENVIRONMENT_VARIABLE_DATA" | tee --append output/file
+sleep 60
+cat output/file
+`[1:]
```
0x2b3bfa0, line 87: My eyes! Can we leave this string where it was? 😅 

I'd also mildly prefer a slightly shorter version like e.g. `test -n "$TEST_GPU" && nvidia-smi` or `printenv TEST_GPU && nvidia-smi` instead of a conditional spanning three lines. 💅🏼

.github/workflows/smoke.yml
```diff
@@ -17,6 +18,14 @@ jobs:
...
     steps:
       - run: echo ✓
+  set_env_pr_test:
+    run: |
+      echo "SMOKE_TEST_TARGET='default'" >> $GITHUB_ENV
+    if: github.event_name == 'pull_request_target'
+  set_env_gpu_test:
+    run: |
+      echo "SMOKE_TEST_TARGET='gpu'" >> $GITHUB_ENV
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```
0x2b3bfa0, line 28: I'd even specify the following in the top level of the YAML document instead of writing those code blocks:

```yaml
env:
  SMOKE_TEST_TARGET: ${{ github.event_name == 'pull_request_target' && 'default' || 'gpu' }}
```

0x2b3bfa0:
Note that `k8s` tests will always run on a GPU-enabled cluster; to avoid this, you may want to modify the following line:

https://github.com/iterative/terraform-provider-iterative/blob/d587a7ff720559349c42648b4a0fd9e935d3174e/.github/workflows/smoke.yml#L109

See the collapsible at https://registry.terraform.io/providers/iterative/iterative/latest/docs/guides/testing-kubernetes#creating-a-test-cluster for more information.

```
az aks create \
  --resource-group testKubernetesResourceGroup \
  --name testKubernetesCluster \
  --node-vm-size Standard_A2_v2 \
  --node-count 1
```

0x2b3bfa0:
Unrelated wish: please increase the `timeout-minutes` for the `text-k8s` job; [30 is not enough](https://github.com/iterative/terraform-provider-iterative/actions/runs/3079737233/jobs/4976341136). 🙏🏼

task/task_smoke_test.go
tasdomas: Do we have a newline quota for the codebase?

0x2b3bfa0: No, we don't, although we don't get paid by the number of lines we write either. 😂 

There is a [natural tendency for compactness](https://github.com/iterative/terraform-provider-iterative/pull/443#pullrequestreview-910911048) in the team you may follow or not. In terms of code duplication there are stronger preferences, but things like this are up to you.

0x2b3bfa0: What about `if test -v TEST_GPU; then nvidia-smi; fi` in a single line? 😼🧶

task/task_smoke_test.go
tasdomas: Ok, added, `quick`, `quick+spot`, `gpu` and `gpu+spot` as test targets.

.github/workflows/smoke.yml
```diff
@@ -5,10 +5,24 @@ on:
...
+        options:
+          - quick
+          - quick+spot
+          - gpu
+          - gpu+spot
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
+env:
+ SMOKE_TEST_TARGET: ${{ github.event_name == 'pull_request_target' && 'quick' || inputs.test_target }}
```
0x2b3bfa0, line 25:
```suggestion
 SMOKE_TEST_TARGET: ${{ github.event_name != 'pull_request_target' && 'quick' || inputs.test_target || 'gpu+spot' }}
```

Matbe make `gpu+spot` the default for the `on: schedule` daily tests?

.github/workflows/smoke.yml
```diff
@@ -91,16 +109,23 @@ jobs:
...
           --node-count=1 \
           --aks-custom-headers=UseGPUDedicatedVHD=true \
           --generate-ssh-keys
+    - id: cluster
+      if: ${{ !contains(env.SMOKE_TEST_TARGET, 'gpu') }}
+      run: |
+        az aks create \
+          --resource-group="tpiSmokeTestCluster$GITHUB_RUN_ID" \
+          --name="tpiSmokeTestCluster$GITHUB_RUN_ID" \
+          --node-vm-size=Standard_NC6 \
```
0x2b3bfa0, line 125: `Standard_NC6` is a GPU-enabled VM size. Did you mean e.g. `Standard_A2_v2` as per https://github.com/iterative/terraform-provider-iterative/pull/673#issuecomment-1249515793?

task/task_smoke_test.go
```diff
@@ -1,9 +1,10 @@
 //go:build smoke
 
-package task
+package task_test
```
0x2b3bfa0, line 3: Thanks! Black-box testing is better at this level. So, we make sure there is no dependency on private API.

task/task_smoke_test.go
```diff
@@ -40,9 +89,17 @@ func TestTaskSmoke(t *testing.T) {
 	}
 
 	if testName == "" {
-		testName = "smoke test"
+		testName = fmt.Sprintf("smoke test: %s", targetName)
```
0x2b3bfa0, line 92: Note that `testName` is fully deterministic on purpose, so we don't inadvertently litter our test cloud accounts with flotsam resources uniwittingly.

task/task_smoke_test.go
```diff
@@ -40,9 +89,17 @@ func TestTaskSmoke(t *testing.T) {
...
 	}
-
+	script := `
+          #!/bin/sh -e
+          [ -n "$TEST_GPU" ] && nvidia-smi
```
0x2b3bfa0, line 96: Even [subjectively] better?
```suggestion
          test -v TEST_GPU && nvidia-smi
```
or
```suggestion
          [ -v TEST_GPU ] && nvidia-smi
```

task/task_smoke_test.go
```diff
@@ -40,9 +89,17 @@ func TestTaskSmoke(t *testing.T) {
...
+          mkdir --parents cache output
+          touch cache/file
+          echo "$ENVIRONMENT_VARIABLE_DATA" | tee --append output/file
+          sleep 60
+          cat output/file
+        `[1:]
```
0x2b3bfa0, line 102: If you really want to keep this script outside, please consider surrounding it with newlines.

```suggestion

	script := `
          #!/bin/sh -e
          [ -n "$TEST_GPU" ] && nvidia-smi
          mkdir --parents cache output
          touch cache/file
          echo "$ENVIRONMENT_VARIABLE_DATA" | tee --append output/file
          sleep 60
          cat output/file
        `[1:]

```

task/task_smoke_test.go
0x2b3bfa0: Moved to https://github.com/iterative/terraform-provider-iterative/pull/673/files#r974235102 & https://github.com/iterative/terraform-provider-iterative/pull/673/files#r974237957. Thanks!

task/task_smoke_test.go
0x2b3bfa0: Thanks!

.github/workflows/smoke.yml
0x2b3bfa0: Thanks!

.github/workflows/smoke.yml
tasdomas: Yes, missed that one - thanks!

task/task_smoke_test.go
tasdomas: Updated. I think it makes it more readable to define the script outside the multiple levels of indentation in the task definition.

task/task_smoke_test.go
```diff
@@ -1,9 +1,10 @@
 //go:build smoke
 
-package task
+package task_test
 
 import (
 	"context"
+	"fmt"
```
0x2b3bfa0, line 7:
```suggestion
```

Imported but not used

task/task_smoke_test.go
```diff
@@ -43,6 +91,16 @@ func TestTaskSmoke(t *testing.T) {
...
 
+	script := `
+          #!/bin/sh -e
+          test -v TEST_GPU && nvidia-smi
+          mkdir --parents cache output
+          touch cache/file
+          echo "$ENVIRONMENT_VARIABLE_DATA" | tee --append output/file
+          sleep 60
+          cat output/file
+        `[1:]
```
0x2b3bfa0, line 102: Not worth addressing before merging, but this works due to the shell interpreting files without shebang: it's ineffective due to the indentation.

PR #693, opened 2022-10-13
Fix `test -v` on smoke tests

0x2b3bfa0:
It turns out that `sh` doesn't understand `test -v`

https://github.com/iterative/terraform-provider-iterative/actions/runs/3239053637/jobs/5307980798#step:5:645


tasdomas (approved):
Thanks!

PR #695, opened 2022-10-13
Use `d.Id()` on Google Cloud machines

### LINKED ISSUE #694: Use `d.Id()` instead of `d.Get("name")` as identifier for Google Cloud
Inconsistent with the other providers and a potential source of issues:

https://github.com/iterative/terraform-provider-iterative/blob/2b46b50aabf249efc19a7f3de6a87a95ef2c5a19/iterative/gcp/provider.go#L34

https://github.com/iterative/terraform-provider-iterative/blob/2b46b50aabf249efc19a7f3de6a87a95ef2c5a19/iterative/gcp/provider.go#L284

Would also eliminate the need for the `name` option when destroying runners:

https://github.com/iterative/terraform-provider-iterative/blob/2b46b50aabf249efc19a7f3de6a87a95ef2c5a19/cmd/leo/destroyrunner/destroyrunner.go#L36

0x2b3bfa0:
Closes #694; releasing this change may cause a hiccup in some Google Cloud users' runners.

### DIFF
--- a/cmd/leo/destroyrunner/destroyrunner.go
+++ b/cmd/leo/destroyrunner/destroyrunner.go
@@ -16,7 +16,6 @@ import (
 )
 
 type Options struct {
-	Name string
 }
 
 func New(cloud *common.Cloud) *cobra.Command {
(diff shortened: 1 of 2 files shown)


tasdomas (approved):
Seeing a significant change like this without it requiring changes to unit tests makes me nervous.

PR #696, opened 2022-10-14
Fix tpi workload.

tasdomas:
After the landing of the tpi command split PR, the machine script was still downloading the terraform-provider-iterative release artifact.
This caused the task to not be run properly.


0x2b3bfa0 (approved):
Thanks! 🤦🏼

PR #700, opened 2022-10-20
Implement `read --follow`

0x2b3bfa0:
* Requirement for #633

### DIFF
--- a/cmd/leo/read/read.go
+++ b/cmd/leo/read/read.go
@@ -68,7 +68,8 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
 		return err
 	}
 
-	for last := 0; ; {
+        var last int
+	for {
 		if err := tsk.Read(ctx); err != nil {
 			return err
 		}
--- a/cmd/leo/read/read.go
+++ b/cmd/leo/read/read.go
@@ -3,7 +3,9 @@ package read
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


dacbd (approved):
Simple and works just fine. Can certainly have a follow-up that is "streaming" logs.

cmd/leo/read/read.go
```diff
@@ -62,39 +68,69 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
 		return err
 	}
 
-	if err := tsk.Read(ctx); err != nil {
-		return err
-	}
+	for last := 0; ; {
```
tasdomas, line 71:
```suggestion
        var last int
	for {
```

cmd/leo/read/read.go
```diff
@@ -62,39 +68,69 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
...
+			return err
+		}
+
+		status, err := o.getStatus(ctx, tsk)
+		if err != nil {
+			return err
+		}
 
-	return nil
+		if delta := strings.Join(logs[last:], "\n"); delta != "" {
```
tasdomas, line 86: This is a really complicated way of saying ` if len(logs) > last {`

cmd/leo/read/read.go
tasdomas: Slightly more idiomatic this way

tasdomas (approved):

cmd/leo/read/read.go
0x2b3bfa0: Thanks! 🙏

cmd/leo/read/read.go
0x2b3bfa0: This accounts for cases where `logs := []string{""}` and an extra newline would be printed in the beginning. Are you sure it's equivalent? 🤔

tasdomas: not sure I understand what you mean - are you trying to work around the case where empty log lines are appended?

0x2b3bfa0: Sort of. It appears that the first returned value of `logs` is no t an empty slice, but a slice containing an empty string.

tasdomas: I see, let's leave it as is then

dacbd (approved):

PR #705, opened 2022-10-27
Remove `cml rerun-workflow`

dacbd:
Closes: #630


0x2b3bfa0 (approved):
:shipit: Breaking but not documented, thus not breaking anything we care about 🎉

casperdcl (commented):
i so happy

PR #707, opened 2022-11-03
Pass spot flag

dacbd:
Pass spot flag to via tfResource for the client (cml runner) to parse


0x2b3bfa0:
Why? Do we still use `tfResource` somewhere? 🤔

dacbd:
So workflows aren't restarted when you terminate an instance, this behavior only makes sense for spot instances.

PR #709, opened 2022-11-04
Server endpoint definition using openAPI and credential parsing middleware

tasdomas:


internal/server/openapi-spec.yaml
```diff
@@ -0,0 +1,197 @@
...
+            - aws
+            # other providers will be added later
+        aws:
+          type: object
+          properties:
+            AccessKeyID:
+              type: string
+            SecretAccessKey:
+              type: string
+            SessionTOken:
```
0x2b3bfa0, line 24:
```suggestion
            SessionToken:
```

internal/server/openapi-spec.yaml
```diff
@@ -0,0 +1,197 @@
...
+  '/task/{id}':
+    get:
+      operationId: getTaskStatus
+      summary: Get task status.
+      parameters:
+        - name: credentials
+          in: header
+          schema:
+            type: string
+            format: byte # byte64 encoded
```
0x2b3bfa0, line 142:
```suggestion
            format: byte # Base64 encoded
```

internal/server/openapi-spec.yaml
```diff
@@ -0,0 +1,197 @@
...
+                $ref: '#/components/schemas/TaskStatus'
+    delete:
+      operationId: destroyTask
+      summary: Deallocate task resources.
+      parameters:
+        - name: credentials
+          in: header
+          schema:
+            type: string
+            format: byte # byte64 encoded
```
0x2b3bfa0, line 165:
```suggestion
            format: byte # Base64 encoded
```

internal/server/server-definition.go
```diff
@@ -0,0 +1,465 @@
+// Package server provides primitives to interact with the openapi HTTP API.
+//
+// Code generated by github.com/deepmap/oapi-codegen version v1.12.2 DO NOT EDIT.
```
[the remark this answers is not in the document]
tasdomas: I'd argue that it makes sense to keep this in version control. By doing that our CI definitions are simpler (no need to install tools to generate the server definition).

task/common/cloud.go
```diff
@@ -31,31 +31,49 @@ const (
...
+			count++
+		}
+	}
+	if count == 0 {
+		return errors.New("empty credentials")
+	}
+	if count > 1 {
+		return errors.New("conflicting credentials")
+	}
+	return nil
```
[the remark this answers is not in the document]
tasdomas: Good point. Updated.

internal/server/server-definition.go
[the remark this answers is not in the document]
0x2b3bfa0: I'd argue[^1] that ultimately we'd at least like our continuous integration workflow to warn us if there is a mismatch between the OpenAPI definitions and the generated code (e.g. if we forget to update them before committing), and that would also require us to install the same tools and perform the same operations.

[^1]: Just because that's something I enjoy, as you may have noticed. 😸

task/common/cloud.go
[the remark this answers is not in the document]
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/709/commits/3b1385b548791bbbc1db5e0854a7c5e85e0cd5a5

task/common/cloud.go
```diff
@@ -30,32 +31,70 @@ const (
 	ProviderK8S Provider = "k8s"
 )
 
+// Credentials define the cloud provider credentials.
 type Credentials struct {
-	AWSCredentials *AWSCredentials
-	GCPCredentials *GCPCredentials
-	AZCredentials  *AZCredentials
-	K8SCredentials *K8SCredentials
+	Provider       Provider        `json:"provider"`
```
0x2b3bfa0, line 36: I know that credentials are processed differently, but it would be nice if we could reuse the `Cloud.Provider` field to validate credentials instead of having a separate `Credentials.Provider` field.

tasdomas: The problem is that we'd have to explicitly pass the provider type separately from the credentials for other endpoints (e.g. to list deployed tasks).

0x2b3bfa0: All the endpoints [should probably] take a provider type, though. 🤔

0x2b3bfa0 (approved):
Approved for landing; remaining comments aren't critical

PR #710, opened 2022-11-04
Add structure for hardcoded credentials

0x2b3bfa0:
To be merged against #708

### DIFF
--- a/task/common/cloud.go
+++ b/task/common/cloud.go
@@ -31,10 +31,10 @@ const (
 )
 
 type Credentials struct {
-	*AWSCredentials
-	*GCPCredentials
-	*AZCredentials
-	*K8SCredentials
+	AWSCredentials *AWSCredentials
+	GCPCredentials *GCPCredentials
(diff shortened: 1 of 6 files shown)


task/common/cloud.go
```diff
@@ -29,6 +30,34 @@ const (
 	ProviderK8S Provider = "k8s"
 )
 
+type Credentials struct {
+	*AWSCredentials
```
tasdomas, line 34: Let's not do this, we gain nothing from using anonymous struct members here.

tasdomas (changes requested):
Let's fix that use of anonymous members and this is good to go.

task/common/cloud.go
0x2b3bfa0: Resolved with https://github.com/iterative/terraform-provider-iterative/pull/710/commits/d95f7d8762972d0ad98ac581b1679909e8d59eb1 😄

tasdomas (approved):

PR #711, opened 2022-11-08
Simple waiting feedback for leo

dacbd:
adds some simple feedback for when users run `leo read --follow` right after a `leo create` and there aren't any logs yet.

### DIFF
--- a/cmd/leo/read/read.go
+++ b/cmd/leo/read/read.go
@@ -70,6 +70,7 @@ func (o *Options) Run(cmd *cobra.Command, args []string, cloud *common.Cloud) er
 
 	var last int
 	firstRun := true
+	waiting := false
 	for {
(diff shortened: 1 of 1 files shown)


tasdomas (approved):

tasdomas:
> Please print to `os.Stderr`

Is that really necessary? In this case the output is purely for UX, `stdout` is perfectly valid.

Granted, leo could use a `quiet` mode that's more automation friendly, but we've got more pressing issues

0x2b3bfa0:
Currently `leo read` only outputs script logs to standard output, and all the other user-oriented messages to standard error. While we still don't provide any stability guarantees, it would be nice to stick to that principle.

0x2b3bfa0 (approved):

PR #716, opened 2022-11-11
Use `logrusctx` instead of `logrus` on `task`

0x2b3bfa0:
The modifications introduced with this pull request can be used to provide detailed operation status, to avoid a a [“The Cloudformation”](https://twitter.com/_euank/status/996513864857481216) experience where users have no insight on the operation progress.

Uses [github.com/0x2b3bfa0/logrusctx](https://github.com/0x2b3bfa0/logrusctx), which is in turn based on [github.com/juju/zaputil/zapctx](https://pkg.go.dev/github.com/juju/zaputil/zapctx), mentioned by @tasdomas in a casual conversation, unsuspecting of my implementation impetus.

### Use case

```go
logger := logrus.WithField("session", sesssionID)
logger.SetOutput(&buf)
ctx = logrusctx.WithLogger(ctx, logger)
tsk.Create(ctx)
// buf will contain the real-time progress of the operation
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/rclone/rclone v1.57.0
 	github.com/sebdah/goldie/v2 v2.5.3
 	github.com/shirou/gopsutil v3.21.11+incompatible
-	github.com/sirupsen/logrus v1.8.1
+	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.5.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.11.0
@@ -49,6 +49,7 @@ require (
 )
 
 require (
+	github.com/0x2b3bfa0/logrusctx v0.1.0
 	github.com/aws/aws-sdk-go-v2/credentials v1.4.3
 	github.com/deepmap/oapi-codegen v1.12.2
 	github.com/go-chi/chi/v5 v5.0.7
--- a/go.sum
+++ b/go.sum
@@ -75,6 +75,8 @@ dmitri.shuralyov.com/html/belt v0.0.0-20180602232347-f7d459c86be0/go.mod h1:JLBr
 dmitri.shuralyov.com/service/change v0.0.0-20181023043359-a85b471d5412/go.mod h1:a1inKt/atXimZ4Mv927x+r7UpyzRUf4emIoiiSC2TN4=
 dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c/go.mod h1:0PRwlb0D6DFvNNtx+9ybjezNCa8XF0xaYcETyp6rHWU=
 git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999/go.mod h1:fPE2ZNJGynbRyZ4dJvy6G277gSllfV2HJqblrnkyeyg=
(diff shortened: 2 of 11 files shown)


dacbd:
I have a couple of questions.

1. why do we need this change?
2. can you show or say how the behavior is being changed as a result of this?
3. Is this something that could be accepted upstream to lorgrus itself?
4. perhaps using a logging library where context is [supported already](https://github.com/rs/zerolog#pass-a-sub-logger-by-context)?

0x2b3bfa0:
1. As stated on the pull the body, to provide [service] users with detailed operation status instead of a laconic and opaque “doing stuff, will take time” message. We don't _need_ this change, but it seems like a nice addition to me.
2. Behavior will stay the same after merging this pull request, but it will be really easy to alter it so logs for individual requests can be stored or sent individually to requesters.
3. No, `logrus` is in maintenance mode and does not accept new features. See the official repository for more information.
4. Yes, that would be nice but would require us to rewrite more code, and we'd risk breaking other unorthodox use cases of the logger like the command-line tool.

See https://github.com/iterative/terraform-provider-iterative/pull/423#pullrequestreview-901140497 for context (pun intended) on the introduction of `logrus` to this project.

dacbd:
> 1. As stated on the pull the body, to provide [service] users with detailed operation status instead of a laconic and opaque “doing stuff, will take time” message. We don't need this change, but it seems like a nice addition to me.

I'm still missing this, I did read the body. To me looks like this lets the logging write out to some buffer that we could then print out to a user via an API call (this is to help enable that correct?), but I see no changes that actually enable this, all is see is `%s/ logger.Info(string)/logger.Info(ctx, string)/gc`  with no other changes

> 2. Behavior will stay the same after merging this pull request, but it will be really easy to alter it so logs for individual requests can be stored or sent individually to requesters.

Should this just go with the changes which utilize the feature?[^1]

> 3. No, logrus is in maintenance mode and does not accept new features. See the official repository for more information.

More for validity Q4?

> 4. Yes, that would be nice but would require us to rewrite more code, and we'd risk breaking other unorthodox use cases of the logger like the command-line tool.

Maybe I'm missing some hacky-ness somewhere but I think that swapping out the logger should be the easiest update/rewrite/refactor that we could do for this codebase?

[^1]: see response to Q1

tasdomas:
I think this is a step in the right direction. The internal task package is going to be used (hopefully) in several different contexts: in the CLI tool, in TPI and in the server. These all have different logging requirements. By adding `context` support to logrus we will be able to use different log writers in different setups.

tasdomas (approved):

0x2b3bfa0:
> _I'm still missing this, I did read the body. To me looks like this lets the logging write out to some buffer that we could then print out to a user via an API call (this is to help enable that correct?), but I see no changes that actually enable this, all is see is `%s/ logger.Info(string)/logger.Info(ctx, string)/gc` with no other changes_

Yes, this pull request just _enables_ the possibility of injecting a custom logger into the context, but doesn't introduce any example.

> Should this just go with the changes which utilize the feature?[1](#user-content-fn-1-5d8692020cf939b2aceddffc78f9c429)

I would explicily like to separate both changes for pull requests like https://github.com/iterative/terraform-provider-iterative/pull/716 (this) and https://github.com/iterative/terraform-provider-iterative/pull/710, so they can be easily cherry-picked to the main branch and separated from commits related to the LEO server implementation.

> Maybe I'm missing some hacky-ness somewhere but I think that swapping out the logger should be the easiest update/rewrite/refactor that we could do for this codebase?

Sure! What could possibly go wrong? 😅

PR #718, opened 2022-11-14
Helper machine type parity with task

dacbd:
ref: https://github.com/iterative/cml/issues/1145

#### Task helpers:
https://github.com/iterative/terraform-provider-iterative/blob/fd4a94041fcf5bab312699342fdbed39ae8ebcb9/task/aws/resources/resource_launch_template.go#L61-L73
https://github.com/iterative/terraform-provider-iterative/blob/fd4a94041fcf5bab312699342fdbed39ae8ebcb9/task/az/resources/resource_virtual_machine_scale_set.go#L111-L123
https://github.com/iterative/terraform-provider-iterative/blob/fd4a94041fcf5bab312699342fdbed39ae8ebcb9/task/gcp/resources/resource_instance_template.go#L72-L84

### DIFF
--- a/iterative/aws/provider.go
+++ b/iterative/aws/provider.go
@@ -486,17 +486,19 @@ func GetAvailabilityZone(region string) string {
 }
 
 func getInstanceType(instanceType string, instanceGPU string) string {
-	instanceTypes := make(map[string]string)
-	instanceTypes["m"] = "m5.2xlarge"
-	instanceTypes["l"] = "m5.8xlarge"
-	instanceTypes["xl"] = "m5.16xlarge"
-	instanceTypes["m+k80"] = "p2.xlarge"
-	instanceTypes["l+k80"] = "p2.8xlarge"
-	instanceTypes["xl+k80"] = "p2.16xlarge"
-	instanceTypes["m+v100"] = "p3.xlarge"
-	instanceTypes["l+v100"] = "p3.8xlarge"
-	instanceTypes["xl+v100"] = "p3.16xlarge"
-
+	instanceTypes := map[string]string{
+		"s":       "t2.micro",
+		"m":       "m5.2xlarge",
+		"l":       "m5.8xlarge",
+		"xl":      "m5.16xlarge",
+		"m+t4":    "g4dn.xlarge",
+		"m+k80":   "p2.xlarge",
+		"l+k80":   "p2.8xlarge",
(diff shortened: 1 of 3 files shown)


0x2b3bfa0 (approved):

iterative/aws/provider.go
```diff
@@ -486,17 +486,19 @@ func GetAvailabilityZone(region string) string {
...
-	instanceTypes["l"] = "m5.8xlarge"
-	instanceTypes["xl"] = "m5.16xlarge"
-	instanceTypes["m+k80"] = "p2.xlarge"
-	instanceTypes["l+k80"] = "p2.8xlarge"
-	instanceTypes["xl+k80"] = "p2.16xlarge"
-	instanceTypes["m+v100"] = "p3.xlarge"
-	instanceTypes["l+v100"] = "p3.8xlarge"
-	instanceTypes["xl+v100"] = "p3.16xlarge"
-
+	instanceTypes := map[string]string{
```
tasdomas, line 489: This will be allocated on every run of this function. Granted, the penalty is small, but I'd probably define this map at the package level.

tasdomas (changes requested):
I'm not a fan of the `map[string]map[string]map[string]string` contraption

iterative/aws/provider.go
dacbd: I had several PRs ago suggested these all be unified and a const but was overruled. (previously this was about regions, but I think the point stands: https://github.com/iterative/terraform-provider-iterative/pull/344)

So again I agree but I think we should resolve this as a separate tech debt PR/discussion.

dacbd:
@tasdomas If you are good with my comments I'll open a separate issue for these.

iterative/gcp/provider.go
```diff
@@ -439,6 +439,15 @@ func getRegion(region string) string {
 
 func getInstanceType(instanceType string, instanceGPU string) (map[string]map[string]string, error) {
 	instanceTypes := make(map[string]map[string]map[string]string)
```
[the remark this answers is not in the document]
dacbd: https://github.com/iterative/terraform-provider-iterative/issues/719

iterative/aws/provider.go
dacbd: https://github.com/iterative/terraform-provider-iterative/issues/720

PR #726, opened 2022-12-01
Improve `machine` identifier format

0x2b3bfa0:
The only ™️ way of doing it

- [ ] depends on https://github.com/iterative/cml/pull/1265

PR #727, opened 2022-12-01
Identifiers with configurable prefixes

tasdomas:
This is needed to be able to differentiate between tasks allocated using leo server and tpi/leo cli.

### DIFF
--- a/iterative/aws/provider.go
+++ b/iterative/aws/provider.go
@@ -46,7 +46,7 @@ func ResourceMachineCreate(ctx context.Context, d *schema.ResourceData, m interf
 	availabilityZone := GetAvailabilityZone(d.Get("region").(string))
 
 	metadata := map[string]string{
-		"Name": d.Get("name").(string),
+		"Name": d.Id(),
 		"Id":   d.Id(),
 	}
 	for key, value := range d.Get("metadata").(map[string]interface{}) {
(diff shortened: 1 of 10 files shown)


0x2b3bfa0:
Also useful for #726 and `cml-` prefixes, but I would argue that there is no need to differentiate `leo-server` from the `leo` command-line tool.

0x2b3bfa0 (commented):
Why enforce the 3 character length requirement? Alternatively, can we just have some `const` prefix values to choose from? E.g. 

```go
type prefix string

const (
    prefixTPI prefix = "tpi"
    prefixLEO prefix = "leo"
    prefixCML prefix = "cml"
)
```

tasdomas:
> Why enforce the 3 character length requirement? Alternatively, can we just have some `const` prefix values to choose from? E.g.
> 
> ```go
> type prefix string
> 
> const (
>     prefixTPI prefix = "tpi"
>     prefixLEO prefix = "leo"
>     prefixCML prefix = "cml"
> )
> ```

I don't see a big difference - keeping it a simple string seems slightly more extendable.

0x2b3bfa0 (approved):
To my mind, the “flexibility” provided by acepting arbitrary prefixes is a bit of a footcannon, but... 🤷🏼‍♂️

PR #732, opened 2023-01-25
Dummy test and lint jobs for ignored paths.

tasdomas:
The `test` and `lint` jobs in test.yml ignore changes to `docs/*` and
`**.md`. Since those checks are required, this results in blocked PRs.

This adds dummy tests to satisfy the required check statuses.

See #731.


0x2b3bfa0 (approved):
Thanks! ❤️ I've code-golfed a bit the workflow, because the officially recommended workflow was too verbose; feel free to revert the last commit.

tasdomas:
Thanks for changes - I think we're close to a code hole in one :)

PR #742, opened 2023-03-01
Port `task` instance types to `runner`

0x2b3bfa0:
Closes https://github.com/iterative/cml/issues/1361

### DIFF
--- a/iterative/kubernetes/provider.go
+++ b/iterative/kubernetes/provider.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"regexp"
 	"strconv"
 	"time"
 
--- a/iterative/kubernetes/provider.go
+++ b/iterative/kubernetes/provider.go
@@ -391,9 +391,13 @@ func getInstanceType(instanceType string, instanceGPU string) (map[string]map[st
 		"memory": {"amount": "512Gi"},
 	}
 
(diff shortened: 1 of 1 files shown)


iterative/kubernetes/provider.go
```diff
@@ -408,6 +413,23 @@ func getInstanceType(instanceType string, instanceGPU string) (map[string]map[st
 		}, nil
 	}
 
+	
+	if match := regexp.MustCompile(`^(\d+)-(\d+)(?:\+([^*]+)\*([1-9]\d*))?$`).FindStringSubmatch(size); match != nil {
```
dacbd, line 417: this was pulled from the task code right? can we easily abstract it out so doesn't exist in two places?

0x2b3bfa0: > _this was pulled from the task code right?_

Yes[^1]

> _can we easily abstract it out so doesn't exist in two places?_

No[^2]

[^1]: With a few minor tweaks.
[^2]: Unless you want to review a dozen pull requests.

dacbd: or are the tweaks different enough that it wouldn't matter?

0x2b3bfa0: Apart from the regular expression, there is nothing in common. I'd merge this as-is. 😅

dacbd (approved):

PR #745, opened 2023-03-13
patch: use same env as dvclive

dacbd:
`STUDIO_URL` -> `STUDIO_ENDPOINT`


omesser (approved):
questionable choice but... K

PR #753, opened 2023-05-03
studio helper update

dacbd:
from the latest release of `dvc-studio-client`
- uses `machine` key
- uses updated env vars


0x2b3bfa0 (approved):
Not tested, but makes sense. Somehow related to https://github.com/iterative/dvc-studio-client/pull/29?

